PR #18, opened 2015-07-02
choose your own adventure

LauraWN:
I need to reorder the statements so they aren't continuing to the next set of if statements if they are supposed to end. I'm still not certain what the buffer() code in the solution is doing even after looking online.

completeness: 4
comfort: 4


RobertAKARobin:
Ask about this in your one-on-one, or in class if there's an opportunity!

PR #21, opened 2015-07-02
homework

lscottcook:
Comfort_level:  1
Completeness:  1

I starteded to try to make an adventure game but I was spending a lot of time trying to create the story for the game and was not getting anywhere.  So I tried to use the same concepts using a different approach but still struggled


RobertAKARobin:
I don't believe you that this is a 1 on completeness!

PR #34, opened 2015-07-04
Choose Your Own Adventure

silviascript:
https://github.com/siltastic/choose_your_own_adventure_js

Comfort Level: 4/5
Completeness: 5/5

I believe I met all the requirements. I decided to base the story from a children's book that I adored as a child. It has a different setup than the provided example. I didn't finish the full story, but I definitely have a better understanding of the if and switch statements. I feel more comfortable about them and the hierarchy in JavaScript. If I have time this weekend, I will finish the full story.

I'm using a lot of if statements and I need to simplify the structure. I've searched for articles to refactor nested if statements, but I haven't found many solutions. Is using the ternary operator a better approach?


RobertAKARobin:
Not really, because it makes things much less readable. An exercise of this kind is going to have a lot of if/else. And that's OK!

PR #38, opened 2015-07-05
JS Oregon Trail Variant Added.

karldonus:
Comfort_Level: 4
Completeness: 4

Observations: 
My attempts in the integer portion was less than satisfying. Theoretically, speaking...should this line of code run? 
while (typeof(boilTime) != 'number') {yadah yadah}
Specifically, I was attempting to run that while loop to prevent a visitor from entering anything other than a number. It just wasn't working despite however many times I tried to play around with it. Really disappointed in myself that I had to leave it out. 

Overall, I thought I made good use of the while loops to force the visitor to only give valid responses....
HOWEVER, this caused HUGE problems down the road. With so many "inception"-like while loops running, it made debugging an absolute nightmare. On the positive side, I learned a better method of debugging. 

I'd love to learn some best-practices in terms of debugging. i.e. What approaches is best?

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -5,6 +5,7 @@
   <title>Document</title>
 </head>
 <body>
+  OREGON TRAIL
   <script src="script.js"></script>
 </body>
 </html>
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,130 @@
 // put code here!
+/*# Choose Your Own Adventure!
+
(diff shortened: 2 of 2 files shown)


RobertAKARobin:
Sure, that looks like it should run. I think the problem is `prompt` always returns a string, I believe.

Ask about debugging in your one-on-one!

PR #41, opened 2015-07-05
Updated script.js - Erica Andersen's Presidential Program

ericacomfort:
Comfort: 3.5/5
Completeness: 5/5

I still need time and practice to fully understand the use of while loops. I was able to get this to run with all 'if/else if/else' conditions. I adjusted them to while loops but I don't think I utilized them properly.


RobertAKARobin:
Bring this up in your one-on-one!

PR #48, opened 2015-07-06
Haunted House

isabellabrookes:
comfort level = 4/5 (quite formulaic once I got in the hang of it).
completion = 3/5 (I got quite stuck on the tic tac toe exercise).

I tried to call a function within in a function like so:

```
if (enterHouse == "y"){
  alert("You get out of the car and run up the hill. \n\nYou miss the sign that say 'Do Not Enter'")
  var knockOnDoor = prompt("Do you want to knock on the door or try the handle? (knock/handle)");
  knockOnDoor1 = knockOnDoor.toLowerCase();
}
//etc etc etc
if (knockOnDoor1 == 'knock'){
  alert("Nobody answers.\n\nTry opening the door yourself.");
  knockOnDoor();
}
```

I was aiming to "loop" back. I assume that the knockOnDoor() returns the input, how would I go about returning the prompt again?

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -2,9 +2,14 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>Document</title>
+  <title>Choose Your Own Adventure</title>
+  <script src='http://use.edgefonts.net/nosifer.js'></script>
+  <script src="script.js"></script>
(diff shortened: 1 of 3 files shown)


RobertAKARobin:
Try putting the prompt within a function, like so:

```
function knockOnDoor(){
    var result = prompt("Yo");
    return result;
}
```

PR #49, opened 2015-07-06
console and html output in this script

MikeDulik:
comfort: 3
completeness: 4

This wasn't a tough assignment, but I spent about 10 hours troubleshooting node vs nodejs in linux. So I learned all kinds of stuff =)


RobertAKARobin:
Sounds like... fun?

PR #56, opened 2015-10-14
andy hw

andrewsunglaekim:
comfort: 4
completeness: 1

Boy andy, this was such a fun hw assignment.... but i was really confused about this one part


andrewsunglaekim:
well this is awkward, do better

PR #60, opened 2015-10-15
Daniel Alexander HW Choose Adventure game

dan-ator:
comfort: 4.5
completeness: 5

comments: I had some trouble because of what I wanted to do as far as looping back, but in the end decided to keep it straight forward and linear.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,135 @@
-// put code here!
+var buffer = function(){
+  console.log("***********************")
+}
+
+console.log("You need to get into the non-descript gray building \n to turn in your form. You are greeted by a security guard.")
+buffer()
+var theGuard = prompt("Please empty your pockets and wait to be frisked.  Do you comply? (please enter 'y' or 'n')") === 'y'
+
+if (theGuard) {
+  console.log("You empty your pockets and are frisked.  The guard finds a switchblade in your pocket.")
+buffer()
+  var switchBlade = prompt("Do you attempt to use the blade on the guard? (please enter 'y' or 'n')");
+
+    if (switchBlade === 'y') {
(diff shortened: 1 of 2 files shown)


starter/script.js
```diff
@@ -1 +1,135 @@
...
+  console.log("***********************")
+}
+
+console.log("You need to get into the non-descript gray building \n to turn in your form. You are greeted by a security guard.")
+buffer()
+var theGuard = prompt("Please empty your pockets and wait to be frisked.  Do you comply? (please enter 'y' or 'n')") === 'y'
+
+if (theGuard) {
+  console.log("You empty your pockets and are frisked.  The guard finds a switchblade in your pocket.")
+buffer()
```
adambray, line 11: make sure to use correct indentation! this should be in one level.

starter/script.js
```diff
@@ -1 +1,135 @@
...
+      console.log("You take a lunging stab at the guard. You miss. He shoots you. You die. You have failed The Bureaucracy.");
+      buffer()
+    }
+    if (switchBlade === 'n') {
+      console.log("The guard confiscates your knife and allows you to enter. You proceed to the elevator.");
+      console.log("You are greeted at the elevator by a tiny old man.");
+      buffer()
+      var oldMan = prompt("Please tell the old man what floor you wish to go to (number 1-100)");
+
+        if (oldMan <= 12) {
```
adambray, line 24: this line shouldn't be indented the extra level, it should be at the same level as the previous line.

adambray:
Hah, this is great! Really captures the soul-sucking experience of being in a bureaucracy.

Without an example, I can't give much feedback on looping, except to say that usually you'd want a while look for continuous looping:

``` js
# arbitrary example
while(form != 'TM-4') {
  # whatever code you want to keep looping
}
```

PR #62, opened 2015-10-15
Kain: Choose Your Own Adventure HW

elisekain:
Comfort Level: 4
Completeness: 5

This game can be played without opening up the console (it uses alerts and document.write to print to the index.html browser window). 

I initially moved the script to the head, but then realized I wanted the title on the page to render first and then run the script, so I moved it back to right before the closing body tag.

When I ran it through JSLint, it said that parseInt was missing the radix parameter, so that is added as suggested. I've used parseInt before without adding a radix...so I'm not sure if this is "required".

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -3,8 +3,9 @@
 <head>
   <meta charset="UTF-8">
   <title>Document</title>
+  <script src="script.js"></script>
 </head>
 <body>
-  <script src="script.js"></script>
+  <h1>The Mystery of the Lost Locket</h1>
 </body>
 </html>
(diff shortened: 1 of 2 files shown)


jsm13:
:+1:  Really awesome job on this -- abstracting out all of the page writing to functions is really clever, makes the code read a lot cleaner and makes it easier to work with!

Recognizing where to include scripts is a great catch and something we'll talk more about in class. 

Also really impressed you put the script through JSlint. It's a great practice to be in and you'll learn a ton from it.

I actually didn't know about the radix being required. Apparently it is best practice to be explicit for the sake of others (or future you) (speaking of, nice pseudocoding and comments) and some cases when parseInt is implemented differently.

PR #65, opened 2015-10-15
Tyler Crosse's choose your own adventure js

tylercrosse:


starter/script.js
```diff
@@ -1 +1,56 @@
-// put code here!
+//=== Program wide functions ===
+var badResponse = function() {
```
amaseda, line 2: Very nice! Keeping that code DRY by condensing repeated lines of code in a function.

starter/script.js
```diff
@@ -1 +1,56 @@
...
+  console.log("You put in something besides a correct response, try again.");
+};
+var response = 0;
+
+//=== Main Story logic ===
```
amaseda, line 7: Good commenting!

starter/script.js
```diff
@@ -1 +1,56 @@
-// put code here!
+//=== Program wide functions ===
+var badResponse = function() {
+  console.log("You put in something besides a correct response, try again.");
+};
+var response = 0;
```
amaseda, line 5: Do you need to initialize this as 0? You can initialize when saving the value of the initial prompt on Line 9.

amaseda:
Great job. Really happy that you created a function to use in your `else` blocks. Encourage you to think about how you could implement reset functionality here so that when the user loses the story automatically begins again.

amaseda:
:+1:

PR #71, opened 2015-10-15
CYOA Complete

aoave:


aoave:
Hw Choose your Own Adventure

Comfortness 3
Completeness 3

adambray:
Hi AO!

This is a good start, but here's some suggestions:
1. Instead of one long line of if statements, it would be better if you used nested `if` statements to better represent the various paths.
2. Related to number 1, it's best not to re-use the same variable for so many different prompts. Instead, I'd make new variables to represent each new question's response: 
   `var hulkResponse = prompt("hulk smash or throw?");`

-Adam

PR #72, opened 2015-10-15
Choose Adventure HW

matiamin:
Hi Andy -

My code isn't executing - it's showing me some error. I'll be getting back to it to fix it, but wanted to share whatever I've so far. Thanks much!

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,67 @@
-// put code here!
+console.log("You might win $100k.");
+var lotteryDecision = prompt("You've $10 in your pocket. You either buy('b') or not ('n') buy a $100k lottery.");
+if(lotteryDecision == "n"){
+  console.log("You return home to find out that your neighbor won the lottery");
+  var partyDecision = ("Your neighbor wins the lottery instead and invites you to his party. You either go ('g'), don't ('d') go, or breakinto ('b') into his house and steall all his money");
+  if(partyDecision == "g"){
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,67 @@
...
+  console.log("You're too sad and decide to stay home");
+  var breakingBad = prompt("You've a crazy idea floating in your head. You either break ('b') into your neighbor's house the next day and steal his money or watch breakingbad ('w') instead for inspiration");
+}
+if(breakingBad == "b"){
+  console.log("He sees you sneaking into his backyard and shoots you in the head. You die.");
+}
+else if(breakingBad == "w"){
+  console.log("You're too sad and die from heart attack.");
+}
+}
```
adambray, line 33: this looks like an extra `}`

adambray:
Hey Mati,

In general, you're on the right track, but I think you'll benefit from a few strategies to prevent getting stuck in the future:
1. Ask for help sooner!
2. Be meticulous about proper structuring of your code, and indentation. This helps find errors like the one here, which is an extra curly brace.
3. If you have code that isn't running, look at the error message(s) in the chrome console. It'll give you a generally helpful message, or at least a line number where the error occurred. You then know to look at that line (or lines right above it) to find the source of the error.
4. If all else fails, try cutting out code (maybe moving it to a temporary file) until it starts working again, then re-add the code back in. Wherever it breaks tells you where the error is.

PR #75, opened 2015-10-15
choose_your_own_adventure Alice Frumin

alifrumin:
Completeness: 4 (did not do parse interger thing)
Comfort: 3.5 I relied heavily on the solution but I feel like I learned a lot about nesting and the general layout
:)


ebirving:
👍

PR #76, opened 2015-10-15
Thom's scary adventure code

ghost:
confidence - 4
completeness - 4.5

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,40 @@
-// put code here!
+
+
+var scared = prompt("It's Halloween.  Are you scared?");
+  if (scared === 'yes') {
+    var candy = prompt("Okay, let's go trick or treating instead. Do you like candy?")
+      if (candy === 'yes') {
+          var pieces=prompt("Great! How many pieces can you eat?")
+            if (pieces > 10) {
+                  prompt("Stomache-ache city! Happy Halloween!")
+              } else if (pieces > 5){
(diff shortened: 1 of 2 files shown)


starter/script.js
```diff
@@ -1 +1,40 @@
...
+              var zombieCount = prompt("You should! Turn around verrry slowly.  How many do you see?")
+              while (zombieCount < 50) {
+                zombieCount = prompt("Now how many?!");
+                zombieCount ++;
+              }
+              alert('Zombies are all around.  You are dead!')
+          };
+
+    }
+  };
```
jshawl, line 40: can you think of a way to start over if the user answers neither "no" nor "yes"?

starter/script.js
```diff
@@ -1 +1,40 @@
-// put code here!
+
+
+var scared = prompt("It's Halloween.  Are you scared?");
+  if (scared === 'yes') {
+    var candy = prompt("Okay, let's go trick or treating instead. Do you like candy?")
+      if (candy === 'yes') {
+          var pieces=prompt("Great! How many pieces can you eat?")
```
jshawl, line 7: Your code looks good! Only code inside curly braces needs to be indented:

``` js
var scared = prompt("It's Halloween.  Are you scared?");
if (scared === 'yes') {
  var candy = prompt("Okay, let's go trick or treating instead. Do you like candy?")
  if (candy === 'yes') {
    var ...
```

jshawl:
:+1:

PR #82, opened 2015-10-16
Alea Q

ghost:
comfort: 3
complete 75% 

I'm having a really hard time absorbing all of this javascript information without real life examples in the classroom.


mattscilipoti:
:+1:

Suggestion: Instead of indicating your progression (final) in the commit message, just indicate what you performed (i.e. html and handles first user choice.)

PR #84, opened 2015-10-16
choose your own adventure hw

lewis2ba:
completeness: 5
comfort: 3/5 - would have liked to do some error handling if the user typed in a weird response, but i did not know how to do this to make it ask the same question again.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,72 @@
-// put code here!
+var counter = 2;
+while(counter>=-1){
+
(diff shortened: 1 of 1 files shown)


RobertAKARobin:
If you want to do REAL error handling, look up Javascript's `try` and `catch` statements. But for now, you could try something like:

```
var input;
while(["apple", "orange", "banana"].indexOf(input) < 0){
  input = prompt("Please enter a fruit.");
}
alert("The fruit you entered was " + input + "!");
```

PR #85, opened 2015-10-16
ly's finished adventure game

Ly900:
Comfort: 3
Completeness: 5

This code worked when I ran it but wreaked havoc in the validator. I'd like to be able to sit down with someone to go through this to clean it up please!


andrewsunglaekim:
Looks fine, i like your use of case statements, very cool. I wanted worry too much about validators for JS, but if you have any concerns plz post any specific examples here or DM me. :+1:

PR #86, opened 2015-10-16
HW 10/15

KoryJCampbell:
completeness: 3
comfort: 3

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,21 @@
+// Kory J. Campbell Choose Your Own Adventure Game
+
+alert("Welcome to Joe's Pizzeria Delivery! A Text-Based Adventure. Press OK.");
+alert("You are a delivery bicycler who has to deliver pizzas to your customers safely! What will this entail..... Press Ok.");
+
+var userInput = prompt("You walk out the Pizzeria, on your way to your customer and you see a little red-haired boy who ask for a slize of pizza. Do you want to give him a slice of your pizza? [type Y or N]");
+
+if(userInput == "Y") {
+  prompt("The red-haired boy takes your pizza happily...however after he quickly gulps it down he asks for another slice...do you fulfill his greedy wishes? [type Y or N]");
+}
+
+if(userInput == "Y") {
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,21 @@
...
+
+var userInput = prompt("You walk out the Pizzeria, on your way to your customer and you see a little red-haired boy who ask for a slize of pizza. Do you want to give him a slice of your pizza? [type Y or N]");
+
+if(userInput == "Y") {
+  prompt("The red-haired boy takes your pizza happily...however after he quickly gulps it down he asks for another slice...do you fulfill his greedy wishes? [type Y or N]");
+}
+
+if(userInput == "Y") {
+  alert("The red-haired boy eats your slices...and as you lay the pizxa down in preparation to get on your bike, he steals your entire pizza. :( GAMEOVER");
+} else(userInput == "N"); {
```
amaseda, line 14: An `else` block contains code that runs if the conditions outlined in your `if` and `else if` blocks aren't triggered. It is never paired with a specific condition. So it usually looks something like this...

``` js
if ( x === 2 ){
  console.log( "x is 2!" );
} else if ( x === 3 ) {
  console.log( "x is 3!" );
} else {
  console.log( "x is something else!" );
}
```

Also be careful with semi-colon placement. You shouldn't include one after the condition. In this case, your `else` condition should look like this: `else if(userInput == "N"){ ... }`

starter/script.js
```diff
@@ -1 +1,21 @@
+// Kory J. Campbell Choose Your Own Adventure Game
+
+alert("Welcome to Joe's Pizzeria Delivery! A Text-Based Adventure. Press OK.");
+alert("You are a delivery bicycler who has to deliver pizzas to your customers safely! What will this entail..... Press Ok.");
+
+var userInput = prompt("You walk out the Pizzeria, on your way to your customer and you see a little red-haired boy who ask for a slize of pizza. Do you want to give him a slice of your pizza? [type Y or N]");
+
+if(userInput == "Y") {
+  prompt("The red-haired boy takes your pizza happily...however after he quickly gulps it down he asks for another slice...do you fulfill his greedy wishes? [type Y or N]");
```
amaseda, line 9: Doesn't look like you're saving the input value of this prompt.

PR #87, opened 2015-10-16
Initial Commit. DraftGuessWho; DraftFootball

ghost:
Comfort 3/5
Completion 1/5

I tried to make something resembling the guess who game. I eventually realized that what it would require to do it correctly was beyond. I tried a second less logic based one that I THOUGHT I understood much better.  However, it is not doing something peculiar that I cannot diagnose. It seems to be bypassing the   if (QAnswer1 = "A")  on row 19 and is going directly to the true part of that statement. I've left it in a semi-diagnose status so you can see what I tried. Including a forced QAnswer1= "C".

However, I think there were a couple things in there I was happy I tried and was able to make work when I compartmentalized the code. I had some randomization for replayability using Math.random() (though I have a suspicion this is where the error is going to be). And some error handling with a While loop that did some data validation work.


starter/script.js
```diff
@@ -1 +1,44 @@
-// put code here!
+
+confirm("Are you ready to play fictional football?")
+var QAnswer1
+var QAnswer = prompt("The crowd roars as the teams take the field. Team captains are sent to the middle of the field. (H)eads or (T)ails?")
+
+while(QAnswer != "H" && QAnswer != "T" && QAnswer != undefined) {
```
jshawl, line 6: You're off to a great start! I think the guess who is a very ambitious idea, but i like it! Maybe you could revisit it during project 1?

The purpose of this exercise was to practice conditionals. Be sure to scope your work down to avoid burnout in the future. e.g.

``` js
var isRaining = confirm("Is it raining outside?")
if( isRaining ){
  console.log("Bring an umbrella.")
} else {
  console.log("No umbrella needed today")
}
```

jshawl:
:+1:

PR #89, opened 2015-10-16
Brendan Mahoney HW

BMahoney08:
Comfort 3/5

Completeness 5/5

I had some trouble with my while loop that encompasses all the code. I was attempting to make the code run the initial prompt again if the user did not input "y" or "n" in the initial prompt. However, when I did this, every time I went through the series of prompts, it would begin again. To work around this I set the variable at the end of each series of prompts equal to either "y" or "n", though this clearly is not the correct way to do it.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -2,10 +2,10 @@
 while(adv !== 'y' || 'n') {
   var adv = prompt("Frodo, would you like to take the ring to Mordor? (Y/N)");
 
-  if(adv == "y") {
+  if(adv === "y") {
     var sam = prompt("Will you take Sam with you? (Y/N)");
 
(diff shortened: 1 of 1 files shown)


RobertAKARobin:
I think the problem is right in your first line, with `while(adv !== 'y' || 'n')`. This makes sense from an English perspective: "while `adv` is not equal to `y` nor `n`". Unfortunately, Javascript doesn't really have "nor": you need to write `while(adv !== 'y' && adv !== 'n')`.

What you wrote works out to: "while `adv` is not equal to "y", OR while "n" is a value." The character "n" is always a value, in the same way that, say, the number 6 is always a value. So this thing will keep repeating.

PR #90, opened 2015-10-16
Add choose your own adventure game-ish thing for JS HW.

mooniker:
Completeness: 4.5 out of 5.
Comfort: 3 out of 5

Tried to do something more complicated and got bogged down so this is a simplified version.


andrewsunglaekim:
Looks great! I like how you had an object to fill out on the onset of your adventure. That was cool. :+1:

PR #93, opened 2015-10-16
Finish Homework Choose Yer Path

brittonwalker:
comfort level: 3.5 
completeness: 4

I was having a lot of trouble with nesting and if else statement in side another. Visually, it was easy to get lost. Learned more carefully looking through the solution though.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,35 @@
-// put code here!
+var buffer = function() {
+  console.log("**************************");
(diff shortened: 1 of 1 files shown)


nolds9:
:+1: Very easy to get lost in the nested conditionals, which is why I recommend psuedo-coding and mapping it out first. Also make sure you are indenting properly.

PR #94, opened 2015-10-16
Pengelly - HW - Choose Adventure

jenpen:
comfort: 3
completeness: 2.5

Issues: 
I would like to include an if statement within the while loop in my second challenge that will kill the user if strength is less than three. Intend to revisit with third challenge based on distance user can jump. 

Completeness is an issue. I overthought the initial layout of the game and had to scale back to reduce complexity. 

Comfort level at onset of the assignment was much lower, but repetitive nature of challenge is very useful. Also found myself understanding the tools to solve with better as I employed.

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>The Sorting Hat</title>
+  <title>The Last Crusade</title>
 </head>
 <body>
   <script src="script.js"></script>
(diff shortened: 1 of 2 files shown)


jenpen:
Issue update: Think I have found possible solution for while loop issue. Logging progress here and continuing to work on it.

andrewsunglaekim:
Yeah, complexity of this problem gets a little out of control when your first option is more than 2 because it just goes up exponentially. Try using the `break` keyword to get an explicit break in your while loop

PR #96, opened 2015-10-16
first attempt

kaustly:
comfort: 2-3
completeness: totally not complete at all

Comments:
I'll keep working on this! I didn't even get to the calculator. 
I'll start submitting work even if it's unfinished... didn't know we were supposed to.


adambray:
Hey Kelly,

Please do submit stuff no matter how far along!

Couple of bits of feedback:
1. Not sure which of the two js files you were working on, please try to indicate that when it's not obvious. (I looked at both this time)
2. Make sure your files end in the correct extension. One needs a `.js` at the end, this ensures it syntax highlights correctly on github.
3. In your `script.js` file, make sure to indent correctly. A rule of thumb is you should indent one level any time you enter a new `{`, and de-indent anytime you close the `}`. You can also choose auto-indent from the atom menus if you're not sure.
4. Related to #3 above, it's a good idea to use blank lines to break up your code between sections an make it more readable.
5. The example in the untitled file makes good use of nesting if statements, but your `script.js` version should do the same to meet the requirements.

Overall, you're on the right path, just keep putting in the effort and you should be fine. If you update this and want more feedback, let me know!

-Adam

PR #99, opened 2015-10-16
tim gregg c.y.o.a hw

tomiteo:
I know it's not a very engaging adventure or exactly to specifications, but it works, it's readable, and I think it preserves the intention of the assignment

comfort 4
compltnss 4

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,39 @@
-// put code here!
+var userInput;
+var suchen = function(message){
+  userInput=prompt(message);
+  if (userInput){
+    return userInput;
+  } else {
+    suchen('TRY AGAIN: ' + message);
+  }
+};
+
+var game = function  () {
+  suchen('Hey! What is your name??');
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,39 @@
-// put code here!
+var userInput;
+var suchen = function(message){
+  userInput=prompt(message);
+  if (userInput){
+    return userInput;
+  } else {
+    suchen('TRY AGAIN: ' + message);
```
jshawl, line 7: nice to see recursion here! others implemented using a while loop

starter/script.js
```diff
@@ -1 +1,39 @@
...
+  }
+};
+
+var game = function  () {
+  suchen('Hey! What is your name??');
+  suchen('Hi there, ' + userInput + '. Wanna play a game?? (y|n)');
+  if (userInput=='y'){
```
jshawl, line 14: try to avoid global variables in the future and instead rely on the return value of `suchen()` here

jshawl:
:+1:

PR #100, opened 2015-10-16
create choose_your_own_adventure

pzhang87:
comfort: 5
completion: 4

I feel pretty comfortable with this, but I should've read the instructions a little more carefully so I don't think every branch asks 3 questions.


nolds9:
:+1:

PR #101, opened 2015-10-16
finished HW choose your adventure by Wonsik Jeong

onesick:
Completness 5/5
Comfortable 5/5


jsm13:
:+1: Really enjoyed your game. Be sure to be careful with types. You have a point in your code where you ask the user to put in a code and then it checks it against an integer. Since the prompt always returns a string, the comparison is always false and the player always loses.

PR #104, opened 2015-10-16
adventure story - QT

qadriyyah119:


nolds9:
:+1: Great job on this. The only thing I would add is make sure you are indenting properly. It might help to get in the habit of opening curly brackets on the same line as the conditional, a la:

```
if (condition){
   // code here
}
```

PR #110, opened 2016-02-15
Jeffreyeaton solution

jeffreyeaton76:
Initial alert box interferes with opening the console in a browser. Just clear that alert, open the console and reload the page. 

Handwritten pseudocode is attached below as an image.
img_1942


joe-gz:
:+1:  This is really great Jeff! love the story, but more importantly the code is really clean and well done.  Great work separating everything into functions so you don't have many nested conditionals.
Also, I love the pseudo-coding; keep that up! it will be super helpful moving forward

PR #113, opened 2016-02-16
Adventure Project

nyron:


one.js
```diff
@@ -0,0 +1,32 @@
+var gender = prompt("are you a boy or a girl?");
+
```
nolds9, line 2: you can we control for case sensitivity?

one.js
```diff
@@ -0,0 +1,32 @@
...
+if(gender === "boy" || gender == "girl"){
+    alert("Let's play a game");
+  }
+else{
+  alert("Start Over");
+ document.write("Entered a failed value, try again");
```
nolds9, line 8: in general, would shy away from `writing` anything directly to the document. Instead, would recommend using jQuery or Vanilla JS to `append` an element to the document.

one.js
```diff
@@ -0,0 +1,32 @@
+var gender = prompt("are you a boy or a girl?");
```
nolds9, line 1: Looks like you are re-using some of the same code in this program-how could we use functions to encapsulate our code?

one.js
```diff
@@ -0,0 +1,32 @@
...
+else if(story1 === "badside"){
+  alert("Go to the bottom of the page and pick badside!")
+}
+else if(story1 === "scaryside"){
+  alert("Go to the bottom of the page and pick scaryside!")
+}
+else{
+  document.write("Entered a failed value, try again");
+}
+}
```
nolds9, line 32: what happens upon completing the story? Could call a function that re-hides all the elements.

one.js
```diff
@@ -0,0 +1,32 @@
+var gender = prompt("are you a boy or a girl?");
+
+if(gender === "boy" || gender == "girl"){
+    alert("Let's play a game");
+  }
+else{
+  alert("Start Over");
+ document.write("Entered a failed value, try again");
+}
+
```
nolds9, line 10: assignment is missing a loop, could loop until the right input is entered

nolds9:
:+1:  Good job on this assignment. Please review comments and let me know if you have any questions.

PR #117, opened 2016-02-16
MKaye Initial Choose Path Solution

MaxKaye:
Draft 1

### DIFF
--- a/InitialCommit
+++ b/InitialCommit
@@ -1,0 +1,112 @@
+var sport = prompt("Time to get coaching advice on a sport!");
+
+var WhichOne = prompt("Do you want to learn about tennis or quidditch? (T/Q)\n[T]ennis!\n[Q]uidditch!").toLowerCase;
+
+switch (WhichOne) {
+  case "t":
+  playTennis();
+  break;
+  case "q":
+  playQuidditch();
+  break;
+}
+
+function playTennis() {
+  alert("Awesome! You have chosen a great sport");
+  LeftRight();
+}
+function LeftRight() {
+  var Handedness = prompt("Do you play with your left or right hand? (L/R)\n[L]eft\n[R]ight").toLowerCase;
+}
+switch (Handedness) {
+  case "l":
+  alert("Sweet, that's a big advantage in tennis! I'm going to have to use unique methods though.");
+  Left-handed();
+  break;
+  case "r":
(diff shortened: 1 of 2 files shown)


Pseudocode
```diff
@@ -0,0 +1,38 @@
+What sport would you like to get advice on?
+Tennis or quidditch? (T/Q)
```
nolds9, line 2: love the pseudo-coding, good job mapping out the control flow. One thing that might be helpful to include is thinking about which segment of program cannot be executed until previous sections have completed. This process could help with brainstorming where to include a loop.

InitialCommit
```diff
@@ -0,0 +1,112 @@
+var sport = prompt("Time to get coaching advice on a sport!");
```
nolds9, line 1: Don't think this variable is being used anywhere, might be a better use case for `alert`

nolds9, line 1: Looks like this was saved as the file `InitialCommit`, would recommend saving it as a `.js` file extension so you get the added benefit of syntax highlighting!

InitialCommit
```diff
@@ -0,0 +1,112 @@
...
+
+var WhichOne = prompt("Do you want to learn about tennis or quidditch? (T/Q)\n[T]ennis!\n[Q]uidditch!").toLowerCase;
```
nolds9, line 3: accounting for case-sensitivity :+1:

InitialCommit
```diff
@@ -0,0 +1,112 @@
...
+
+switch (WhichOne) {
```
nolds9, line 5: nice use of a switch statement!

InitialCommit
```diff
@@ -0,0 +1,112 @@
...
+        alert("This is essential in tennis so run and stretch a lot, and eat healthy");
+        break;
+        case 2;
+        alert("Split step right before opponent strikes the ball");
+        break;
+        case 3;
+        alert("Toss the ball high, stay low to the ground, and then spring up upon striking the ball");
+        break;
+        case 4; alert("Uh oh! you may need a different coach..");
+        break;
```
nolds9, line 58: would recommend setting `default`s for your `switch` statements to account for user input not being checked in your cases.

nolds9:
:+1:  Good, well-organized, semantic code. Only thing missing is to incorporate loops, perhaps on initial sport selection? Please review comments and let me know if you have any questions

PR #121, opened 2016-02-16
Keith's CYOA JS

khenry10:


starter/script.js
```diff
@@ -1 +1,113 @@
...
+      // --> Start of NH's first 3rd Level <--
+      if (cokeOrAffair === 'coke') {
+        confirm('Busted.  Your intern got arrested and ratted you out.');
+      } else if (cokeOrAffair === 'affair') {
+        confirm('Brilliant.  Secretary Clinton jumped all over the accusation and has badly bruised Sanders in the polls, while you keep distance yourself, avoid all suspicion, and continue to rise.');
+      } else confirm('You failed to follow simple instrustions.  Reload the page to start over.');
+      // --> End of NH's first 3rd Level <--
+    } else if (moneyOrSabotage === 'm') {
+      for (i = 0; i < oldArray.length; i += 1) {
+        newArray.push(oldArray[i] = prompt('What 3 things should we spend the increased budget on? ' +  (3 - i) + ' more.'));
```
joe-gz, line 59: really small note, to save two characters of text you can just use i++, instead of i+= 1!

starter/script.js
```diff
@@ -1 +1,113 @@
...
+        } else if (resignOrNot === 'n') {
+          alert('You have guts, but the Underwoods now own you')
+        } else null
+        // --> End of Nevada's 2nd 3rd Level <--
+      } else null;
+    // --> End of Nevada 2nd Level <--
+  } else confirm('You failed to follow simple instrustions.  Reload the page to start over.')
+
+/**
+The below "||"" was messing this up.  I wasn't able to figure out why, but but I would like to find out!!
```
joe-gz, line 104: This is a great question! when using the "||", or the "&&" operators, you still need to include the first part of the operation on both sides. So something like:

if (userPrimaryPick === 'Iowa' || userPrimaryPick === 'iowa')

starter/style.css
```diff
@@ -0,0 +1,7 @@
+body {
+  background-image: url('images/house_of_cards.jpg');
```
joe-gz, line 2: :+1:

joe-gz:
:+1: Great job Keith! See the comments i wrote in the code above.  A couple other higher levels things..
just be sure to include your comfort and completeness levels when submitting in the comments for the pull request! It helps give me an idea of where you're at while grading.
Now that you have it working, how would you go about pulling out the nested if statements and using functions instead?

joe-gz:
I should also note, the pseudo-code is great! keep that up, it will really help you out in the long run

PR #122, opened 2016-02-17
charleshwang_solution

ghost:
comfort: 3.5/5
completeness: 5/5

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,58 @@
 // put code here!
+var name = prompt ("What is your name?");
+alert ("Hello," + " " + name + "." + " You are a professional chef and need to prepare a meal for Gordon Ramsey.");
+
+meal ();
+
+function meal () {
+  var mealtype = prompt ("Are you going to make breakfast, lunch, dinner, or dessert?");
+  if (mealtype === "breakfast") {
+    breakfast ();
+  } else if (mealtype === "lunch") {
+    lunch ();
+  } else if (mealtype === "dinner") {
+    dinner ();
+  } else if (mealtype === "dessert") {
+    dessert ();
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,58 @@
 // put code here!
+var name = prompt ("What is your name?");
+alert ("Hello," + " " + name + "." + " You are a professional chef and need to prepare a meal for Gordon Ramsey.");
+
+meal ();
+
+function meal () {
+  var mealtype = prompt ("Are you going to make breakfast, lunch, dinner, or dessert?");
```
beckybeauchamp1, line 8: Instead of naming this variable mealtype, make sure to use JS convention of camelCase. I would instead call it `var mealType`. Not a big deal here, but it's good to get in the habit of using the convention!

starter/script.js
```diff
@@ -1 +1,58 @@
...
+  } else if (mealtype === "lunch") {
+    lunch ();
+  } else if (mealtype === "dinner") {
+    dinner ();
+  } else if (mealtype === "dessert") {
+    dessert ();
+  }
+};
+
+function breakfast () {
```
beckybeauchamp1, line 20: Great use of the while loop here!

beckybeauchamp1:
Hi Charles, 

Great job on Choose your Own Adventure! I love that you separated everything into separate functions, and your choices in naming variables and functions. I made a few additional inline comments, but I really enjoyed your story and the way you broke up your code!

One thing to note is that I did not see your pseudo code included. Especially for an exercise like this, pseudo code is super useful in helping to identify and break up your problem before jumping into code. 

Please let me know if you have any more questions :+1:

PR #125, opened 2016-02-17
Mohamed_Solution

mohamedrchalal:
some bugs still there about 90% finished, no number inputs added :-(

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,74 @@
-// put code here!
+function startGame() {
+var name = prompt('What is your NAME?');
+'hello '+ name + '!';
+var Class = prompt('choose WARRIOR or ROGUE');
+  'you have chosen' + Class + '!';
+  if (Class == 'ROGUE'){
+    banditResponse = prompt('a bandit camp lay ahead do you FIGHT or SNEAK?');
+  }
+  if (banditResponse == 'FIGHT'){
+    death();
+  }
+    else if (banditResponse == 'SNEAK'){
+      riverChoice = prompt('a river is ahead do you SWIM or take the BRIDGE a mile away?');
+  }
+  if (riverChoice == 'BRIDGE'){
+    death();
+  }
+    else if (riverChoice == 'SWIM'){
+      houseChoice = prompt('soaking wet, you see a house. Do you ENTER, or do you PASS it by?');
+    }
+  if (houseChoice == 'PASS'){
+    death();
+  }
+    else if (houseChoice == 'ENTER'){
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,73 @@
-// put code here!
+function startGame() {
+var name = prompt('What is your NAME?');
```
nolds9, line 2: careful of indentation

starter/script.js
```diff
@@ -1 +1,73 @@
...
+'hello '+ name + '!';
```
nolds9, line 3: what about alerting this string, currently this is not shown to the user.

starter/script.js
```diff
@@ -1 +1,73 @@
...
+var Class = prompt('choose WARRIOR or ROGUE');
+  'you have chosen' + Class + '!';
```
nolds9, line 5: could `console.log` or `alert` this string so the user has a visual interface with your program

starter/script.js
```diff
@@ -1 +1,73 @@
...
+  if (Class == 'ROGUE'){
```
nolds9, line 6: how are we controlling for user case sensitivity?

starter/script.js
```diff
@@ -1 +1,73 @@
...
+  }
+  if (banditResponse == 'FIGHT'){
+    death();
+  }
+    else if (banditResponse == 'SNEAK'){
+      riverChoice = prompt('a river is ahead do you SWIM or take the BRIDGE a mile away?');
+  }
+  if (riverChoice == 'BRIDGE'){
+    death();
+  }
```
nolds9, line 17: Would be mindful of indentation in your conditionals as it is easy to lose track of nested control flows.

starter/script.js
```diff
@@ -1 +1,73 @@
...
+      chaliceChoice = prompt('you get to a cave where two chalices lay in front of you, a PLAIN wooden one, and an adorned GOLDEN one. Which of these is the holy grail? the PLAIN or the GOLDEN chalice?');
+    }
+      if (chaliceChoice == 'PLAIN'){
+        'This is not indiana jones you ' + name + '!';
+        death();
+      }
+        else if (chaliceChoice == 'GOLDEN'){
+          warriorWish = prompt ('great magic overwhelms you, a booming voice tells you to make a wish. What is your wish?');
+          death();
+        }
```
nolds9, line 63: An easy low-hanging fruit is to go back and add `else`'s to account for user input that does not match exactly our checks.

nolds9:
:+1:  Great job on this assignment. Please review comments and let me know if you have any questions.

PR #128, opened 2016-02-18
Winnie's Vacation Adventure

winishahines:
comfort level: 4
completeness level: 4

The else statement on lines 47-49 should require the user to go back and choose on of the available options. I couldn't figure out how to do this, any thoughts?

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -17,7 +17,7 @@ if (destination==="F") {
         }
 
   } else if (destination==="I") {
-    alert ("Congrats! Your going to Italy!");
+    alert ("Good choice! Your going to love Italy!");
     var whoIsGoing = prompt ("How many people, including you, will be going?");
 
         if (whoIsGoing < 3) {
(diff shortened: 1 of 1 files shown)


RobertAKARobin:
You could use `while(true)`:

```
while(true){
  var input = prompt("Say hello");
  if(input == "hello"){
    break;
  }else{
    alert("That's not hello!");
  }
}
```

starter/script.js
```diff
@@ -1 +1,69 @@
-// put code here!
+var name = prompt ("What's your name"); alert ("Nice to meet you" + " " + name + "." + " " + "You just won a free vacation!");
+
+var destination = prompt ("Pick a destination: Fiji, Italy, South Africa (F,I,S) ");
+
+if (destination==="F") {
+  alert ("Great! Fiji has excellent beaches.");
+  var whoIsGoing = prompt ("How many people, including you, will be going?");
+
+        if (whoIsGoing < 3) {
```
RobertAKARobin, line 9: Watch your indentation! This says to me that `if (whoIsGoing < 3){` is inside `var whoIsGoing`... which is impossible!

starter/script.js
```diff
@@ -1 +1,69 @@
...
+            alert ("Fabulous! We’ll book you and your guests on the riviera.");
+        }
+        else if (whoIsGoing < 5 ) {
+            alert ("Great! There's an awesome hotel a few blocks from the Leaning Tower of Pisa.");
+        }
+        else {
+            alert ("Woah! That’s more people than we expected. You’ll have to foot some of the bill!");
+        }
+  }
+  else if (destination==="S") {
```
RobertAKARobin, line 33: Take a look at [switch...case](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch). It saves you a lot of trouble when you're writing something like:

```
if x === "a"
else if x === "b"
else if x === "c"
else if x === "d"
...
```

PR #129, opened 2016-02-18
Alexa Klein Adventure Solution

aklein916:
Took me a long time to begin understanding how to write the functions and get it to run. Finally have a better understanding, but had trouble figure out the loops and how to incorporating them. 

completeness: 4 (no loop)
comfort-level: 3.5-- I had ideas to make it more complicated but did not know how to do so. I finally know some things that I don't know!

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,36 @@
-// put code here!
+var welcome = prompt("Welcome to Washington, DC! What is your name?")
+var season = prompt("You'll have a great time here! Are you here in the spring, summer, fall or winter?")
+  if (season === spring || season === summer || season === fall or season === winter) {
+    prompt("There is so much to do in Washington, DC during the" + " " + season. "Do you want to do an activity inside or outside?")
+  }
+  else {
+    alert("Please enter a season. Either spring, summer, fall or winter.")
+  }
+    var place;
+    If (season === spring && place === outside) {
+      prompt("It's the perfect time of year! The weather isn't too hot or humid yet. It's also Cherry Blossom season! Do you want to see the cherry blossoms or check out Eastern Market?")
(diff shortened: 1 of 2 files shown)


starter/script.js
```diff
@@ -1 +1,98 @@
...
+    alert("The view up here is great! Especially in the winter when there are no leaves on the trees. Get there early to get a ticket! And don't forget your camera.");
+  }
+  else {
+    alert("Please enter either 'G' or 'WM' to continue.");
+    sportOrMonument();
+  }
+}
+function end() {
+  alert("Thanks for coming! Enjoy your visit and come again soon!")
+}
```
amaseda, line 98: Good use of functions here. Do think about, however, when it's best to use a function and when it's not. If these functions are not being re-used, is it worth defining them in the first place? Maybe you can just write all that code out in sequential order. Based on how you've set up your conditionals, though, I think using them here was a good call.

starter/script.js
```diff
@@ -1 +1,98 @@
...
+    var place = prompt("Great! Do you prefer to be inside or outside? Enter 'I' for inside or 'O' for outside.");
+    if (place === "O") {
+      boatOrRUn();
+    }
+    else if (place === "I") {
+      museumOrHouse();
+    }
+    else {
+      alert("Please enter either 'I' or 'O'.");
+      location1();
```
amaseda, line 22: I like this recursive way of keeping the prompt going in the case of a bad input. Can you think of other tools at your disposal that could accomplish the same thing?

amaseda:
:+1: Good job with this exercise. Let me know if you have any additional questions/comments. Also happy to sit down and go through how to incorporate a loop into this exercise.

PR #137, opened 2016-02-18
My adventure submission

jblakeman:
My comfort level with this one was ok.  The game is a bit recursion heavy.  Not implementing it that way would've probably saved me some time debugging and made the script less error prone.


joe-gz:
:+1: This is great John! I honestly don't have much to comment on in terms of cleaning it up. If you wanted a bonus challenge, maybe try converting this into one object that still runs the game? Let me know if you have any specific questions

PR #138, opened 2016-02-18
Farishta- Choose your own adventure

LaFrish:
Comfort level 4
Completeness  3.5

Ran out of time but want to add more.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,55 @@
-// put code here!
+alert("You must be 21 to partake in Happy Hour.\nUntil then, enjoy the drinking experience vicariously with this adventure!");
+
+alert("After a long day, it's time to relax with a bevey. You're at your regular drinking hole with friends. Larry, the bartender, asks what you will have to drink.");
+
+var response = prompt("What will be your first drink of the night?"
++ "\n For Beer enter B"
++ "\n For a Cocktail enter C"
++ "\n For a Shot enter S");
+
+if (response === "B") {
+  alert("Enjoy a cold, frosted beer.");
+}
+else if(response === "C") {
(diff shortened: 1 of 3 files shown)


starter/script.js
```diff
@@ -1 +1,55 @@
...
+}
+else if(response === "C") {
+  alert("This cocktail was just what the Doctor ordered");
+}
+else if (response === "S") {
+ alert("SHOTS!SHOTS!SHOTS! Guaranteed fun!");
+}
+
+// If Beer was choosen//
+
```
jshawl, line 21: shouldn't this code be nested under the conditional on line 10?

starter/script.js
```diff
@@ -1 +1,55 @@
...
+var B = prompt("There is a friends' going away gathering at a nearby bar and your friends at the bar are going."
++ "\nAre you going to go to the going away party?"
++ "\n"
++ "\nNo"
++ "\nYes");
+
+if (beer === "No") {
```
jshawl, line 28: can you think of a way to handle user input like "no" instead of only "No"?

starter/stylesheet.css
```diff
@@ -0,0 +1,19 @@
...
+
+    <script>
+    function myFunction() {
+        var drink = prompt("After a long day, it's time to relax with a bevey. You're at your regular drinking hole with friends. Larry, the bartender asks what you will have to drink. Will you start the night off with Beer, a cocktail or a shot, Beer, Drink");
+
+      if (person != null) {
+          document.getElementById("demo").innerHTML =
+          "Hello " + person + "! How are you today?";
+      }
+  }
```
jshawl, line 19: html and js in a css file?!

jshawl:
:+1:

PR #139, opened 2016-02-18
Brendan Magee – Choose Your Own Adventure JS

brendaneamon:
Comfort: 4
Completeness: 5

I believe I hit the requirements, but I would love feedback, as always. Also, please don't judge me too harshly for the subject matter!

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,71 @@
-// put code here!
+var name = prompt("What's your name, buddy?", "Dan");
+alert("Here we are again, " + name + ". " + "You find yourself stumbling out of your favorite watering hole at 11 PM on a Thursday. You notice the familiar glow of a dodgy pool hall immediately to your right. The bar is sparsely populated with mulleted patrons, so you ponder your next choice carefully.");
+
+var boozeMore = function() {
+  var choiceOfVenue = prompt("Do you want to enter the pool hall, or would you prefer to go home and pass out on your rug?", "pool hall or home");
+  if (choiceOfVenue == "home") {
+    return homeFridgeBeers();
+  } else if (choiceOfVenue == "pool hall") {
+    return poolHallDrinks();
+  } else alert("That's not one of the options I gave you, brah!");
+    return boozeMore();
+};
+
+var homeFridgeBeers = function() {
+  var sleepyRoommie = prompt("You come to find your roommate, Chris, sleeping on the couch, with approximately 14 High Life bottles next to him. Do you leave him in peace and go to the fridge, or do you draw on him with a sharpie?", "sharpie or leave alone");
+  if (sleepyRoommie == "leave alone") {
+    alert("Your lame-o roommate drank all the High Life! Looks like it's back to the pool hall . . .")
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,71 @@
...
+alert("Here we are again, " + name + ". " + "You find yourself stumbling out of your favorite watering hole at 11 PM on a Thursday. You notice the familiar glow of a dodgy pool hall immediately to your right. The bar is sparsely populated with mulleted patrons, so you ponder your next choice carefully.");
+
+var boozeMore = function() {
+  var choiceOfVenue = prompt("Do you want to enter the pool hall, or would you prefer to go home and pass out on your rug?", "pool hall or home");
+  if (choiceOfVenue == "home") {
+    return homeFridgeBeers();
+  } else if (choiceOfVenue == "pool hall") {
+    return poolHallDrinks();
+  } else alert("That's not one of the options I gave you, brah!");
+    return boozeMore();
```
amaseda, line 11: :+1: use of recursion, brah!

starter/script.js
```diff
@@ -1 +1,71 @@
...
+        } else return sadWalk();
+    } else if (poolHall == "go to the jukebox") {
+      alert ("You make your way confidently to the TouchTunes jukebox by one of the pool tables, beneath the FC Barcelona poster.");
+      return jukeboxHero();
+    } else if (poolHall == "order a beer") {
+      return howDrunkToGet();
+    } else alert(". . . Seriously? Come on!")
+      return poolHallDrinks();
+};
+
```
amaseda, line 48: :+1: Really like how you organized this assignment using functions.

amaseda:
:+1: Awesome job with this. While functions took care of that loop-like behavior here, encourage you to think about how you could use an actual loop to do this as well. No need to change your code though!

Let me know if you have any additional questions/comments.

starter/script.js
```diff
@@ -1 +1,71 @@
...
+var howDrunkToGet = function() {
+  var howManyDrinks = parseInt(prompt("You belly up to the bar and order some beers. How many will you have?"));
+  while (howManyDrinks <= 6) {
+    alert("Never one to be bashful about lager, you guzzle your " + howManyDrinks + " beer(s) and order more." )
+    howManyDrinks += 2;
+  }
+  alert("Well, you've already been drinking for eight hours, so it's about time to call it a night. The owner of the pool hall appears to appreciate your patronage—he even invites you back to watch the Notre Dame game on Saturday. You head home, pleased with yourself. Another successful night on the town!");
+};
+
+boozeMore();
```
amaseda, line 71: How would you go about dealing with user inputs that **almost** match your `if` and `else if` statements (ex. what if the user submits a correct response but in all caps)?

brendaneamon, line 71: I've done a bit of reading and research, and I'm thinking that I would have to account for such user input using string methods like `toLowerCase()` or `includes()`. That way, an entry in all caps would be converted to lowercase and would still result in the expected outcome. Am I on the right track with this?

PR #140, opened 2016-02-18
Matt K's Solution

mkernsNCR:
comfort: 1
completeness: 1

Comments: I'm having trouble getting started. I don't know how/where to begin.


mattscilipoti:
I see that you were having some trouble here.  Let's meet to discuss and work through.

PR #143, opened 2016-02-18
Anissa solution

iamanissa:
Comfort score: 3.5 out of 5
Completeness score: 3.5 out of 5
Any comments: I was way too ambitious for the time limit. I also, would have liked to figure out how to implement more functions, because I found that my code could have been simplified. Pseudo code saved my life, glad I learned that before I started this homework.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -3,12 +3,109 @@
 //Goal is to get all the makeup, hair, and clothing neccessities for the Princess. And bring
 // them back so that she can go to THE red carpet event.
 
-//Rainbow throwup = lip sticks
-//Phoenix ashes = smokey eyes
-//Mama Troll Buttermilk icecream = foundation
-//Fairy dust = shimmer and highlights
-//Cracken ink = eye liner and mascara
-//Dragon scales = dress
-//Mermaid hair = extentions or weave
-//Medusa hair = necklace
 //Elf workers gift = heels/shoes
+//Unicorn Rainbow throwup = lip sticks
+//Medusa hair = necklace
+//Dragon scales = dress
+//Fairy dust = shimmer and highlights
+//Mermaid hair = extentions or weave
+//Phoenix ashes = smokey eyes
+
+
+
+//Print to screen: The Millinnial Princess Pink Apply Everly Reignbeau aka Princess Pear of Mac'Ephora Lusha Brown Land is getting ready for
+//her first Golden Carpet event.
+//She has summoned you to her chambers. What is your name great warrior?
+//User enters name.
+//Princess Oh 'Users Name', all the things I have are so common, the other Princesses will
+//surely have the same thing on. I need you to find me the rarest of the rare things that
+//will make my beauty stand out above all the others. Will you help.
+//(y)es, gladly, were do I start?,
+      //First go to the elves in the basement, I hear they make shoes.
+//(N)o, unless you fill my pockets with gold,
+      //Fine you heathen, I'll pay. Now go to the elves in the basement, I hear they make shoes.
+// or (B)ut your beauty outshines the others, you could wear a trashbag.
+      //Gross, I would rather die. Please just go to the elves in the basement, I hear they make shoes.
+
+                    //All at elves
+//The elves are working hard
+//You ask for help. One elf steps forward and says
+//They are tired and cannot take on anymore orders. The princesses from around the world
+//Trapped them here. However, if he frees them, they will make one last pair for Princess Pear
+//Choose : (y)es of course,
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,212 @@
...
+
+//Items to get:
+//Elf workers gift = heels/shoes
+//Unicorn Rainbow throwup = lip sticks
+//Medusa hair = necklace
+//Dragon scales = dress
+//Fairy dust = shimmer and highlights
+//Mermaid hair = extentions or weave//Phoenix ashes = smokey eyes
+
+//First I need something to hold items: a dictionary/key:value pair.
```
amaseda, line 14: I like the way you write out your pseudo-code!

starter/script.js
```diff
@@ -1 +1,212 @@
...
+var backPack = {'Unicorn Rainbow Lipstick':0, 'Mermaid Weave':0, 'Dragon Scale Dress': 0,
+  'Medusa necklace':0, 'Stilletoes by ELF': 0, 'SmokeyEye by PHOENIX': 0, 'Fairy Face Shimmer': 0}
+
+function noAnswerLoop(reply){
+  while(reply === ""){
+    var reply = prompt("You must make a choice, " + userName);
+  }
+  return reply;
+};
```
amaseda, line 23: Interesting way of ensuring the user provides a response!

starter/script.js
```diff
@@ -1 +1,212 @@
...
+var userName = prompt("Princess Pear: 'What is your name great warrior?'" , "Type Name.");
+
+//Princess Pear's Mission
+//If I console.log(firstReply); I do get whatever I put in. So the error is in the loop.
+alert("Princess PEAR: \n'Oh," + " "+ userName + " " + "all the things I have are so common, the other Princesses will surely have the same thing on. I need you to find me the rarest of the rare things that will make my beauty stand out above all the others. \n\nWill you help?'");
+var firstReply = prompt("Will you help the princess? \n\nType 'Y' for 'Yes, gladly, where do I start' \nOR 'N' for 'No, unless you fill my pockets with gold' \nOR 'B' for 'But, your beauty outshines the others, you could wear a trashbag.'");
+
+//How will user began journey?
+
+firstReply = noAnswerLoop(firstReply);
```
amaseda, line 39: Can you think of a way to write `noAnswerLoop` (and perhaps rename) so that you **do not** have to ask for an initial prompt and then also call `noAnswerLoop`?

starter/script.js
```diff
@@ -1 +1,212 @@
...
+
+
+if ((firstReply === 'y') || (firstReply === 'Y')){
```
amaseda, line 42: Look into using a method like [`.toLowerCase`](http://www.w3schools.com/jsref/jsref_tolowercase.asp) to shorten your conditional statements.

starter/script.js
```diff
@@ -1 +1,212 @@
...
+//Unicorn Rainbow throwup = lip sticks
+//Medusa hair = necklace
+//Dragon scales = dress
+//Fairy dust = shimmer and highlights
+//Mermaid hair = extentions or weave//Phoenix ashes = smokey eyes
+
+//First I need something to hold items: a dictionary/key:value pair.
+var backPack = {'Unicorn Rainbow Lipstick':0, 'Mermaid Weave':0, 'Dragon Scale Dress': 0,
+  'Medusa necklace':0, 'Stilletoes by ELF': 0, 'SmokeyEye by PHOENIX': 0, 'Fairy Face Shimmer': 0}
+
```
amaseda, line 17: What is your reasoning for using an object to store your inventory? Nothing wrong with the way you have it now, but it might be simpler to initialize `backPack` as an empty array and `push` items to it as the story progresses.

And when you end up implementing that pseudo-code at the end, you can just check the length of the array.

starter/script.js
```diff
@@ -1 +1,212 @@
...
+var guessNum = prompt("This will be easy, so you agree to Hash's game.", "Pick a number between 1 and 20.");
+guessNum = noAnswerLoop(guessNum);
+
+var hashNum = Math.floor((Math.random() * 20) +1);
+
+if(guessNum === hashNum){
+  alert("Thank you for keeping an old Phoenix company. \nThe secret is to take my ashes for the perfect smokey eye eyeshadow.");
+  alert("Hash dies leaving behind his dust. You gain *SmokeyEye by PHOENIX*.");
+  backPack['SmokeyEye by PHOENIX'] = 1;
+}
```
amaseda, line 201: One way to make this more interesting is to add some code that tells the user whether his or her guess is higher or lower than the randomly-generated number.

amaseda:
:+1: Good job with this. Really enjoyed the story! Let me know if you have any remaining questions/comments.

PR #144, opened 2016-02-18
Toni Langley solution

tvcrawley:
completeness: 5
comfortability: 5

comment: I think I spent too much time on this mainly because I wanted it to be able to run correctly.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,298 @@
-// put code here!
+//some kind of loop
+//function intro(){
+  alert("Welcome! You are in for a real treat today.");
+  console.log("Welcome! You are in for a real treat today.");
+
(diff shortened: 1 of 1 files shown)


RobertAKARobin:
Dang, dude. This is HUGE! It looks great! Clearly you're good at getting it to work. Now, I'd focus on tidying up your code. Some thoughts:
- Put all your text together, instead of having it scattered through your code
- Define all your functions together at one place on your code instead of defining new functions throughout your code
- If you're doing something like `if(x == "a"){ ... }else if(x == "b"){ ... }else if(x == "c"){ ... }`, then [switch...case](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) can save you a LOT of typing.

PR #146, opened 2016-02-18
Kim's CYOA HW

kimcabb:


mattscilipoti:
:+1: Looks good.  I especially like the use of functions for ChooseTransport.  I didn't see any number inputs or loops.

PR #147, opened 2016-02-18
LNA1

LNewsome:
Working on defining variables, and sorting out path ways. I copied from an original repository on my Github. Here's the link.
https://github.com/LNewsome/Labyrinth_Adventure

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,93 @@
-// put code here!
+var side = prompt ("Your shirt is atrocious! I'm taking your dog! You'll have to solve my labyrinth. You have 13 mins. Pick your path! Left or Right! ");
+
+if (side == "left") {
+  alert("Boog of Eternal Strentch");
+
+  var boog_direction = prompt ("Oh! I see you're in the Boog of Eternal Stentch. Which way do you want to go. Left, or right. Your choice.");
+  var side;
+      if (side == "left") {
+          alert("My Rock friends help me across!I can go into the Forest.");
+      }
+      else (side == "right");{
+          alert("Oh No! I fell into the Bog of Eternal Stentch!");
+      }
+      var forest_direction = prompt ("I didn't think you'd get this far! You look hungry. Here have a peach?");
+         var eat;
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,93 @@
-// put code here!
+var side = prompt ("Your shirt is atrocious! I'm taking your dog! You'll have to solve my labyrinth. You have 13 mins. Pick your path! Left or Right! ");
+
+if (side == "left") {
+  alert("Boog of Eternal Strentch");
+
+  var boog_direction = prompt ("Oh! I see you're in the Boog of Eternal Stentch. Which way do you want to go. Left, or right. Your choice.");
+  var side;
+      if (side == "left") {
```
mattscilipoti, line 8: You just populated `boog_direction` with my choice, but now your condition is based on `side`.

starter/script.js
```diff
@@ -1 +1,93 @@
...
+          alert("My Rock friends help me across!I can go into the Forest.");
+      }
+      else (side == "right");{
+          alert("Oh No! I fell into the Bog of Eternal Stentch!");
+      }
+      var forest_direction = prompt ("I didn't think you'd get this far! You look hungry. Here have a peach?");
+         var eat;
+             if (eat == "Okay") {
```
mattscilipoti, line 16: Similar issue here.  You want to be using the result of the prompt (`forest_direction`)  or rename that to `eat` (makes more sense to me).

mattscilipoti, line 16: Next time, I recommend you take small steps and try it out along the way.  You have 93 lines of code, which break at line 8.

mattscilipoti:
:+1:  You are really close.  A few tweaks and this will do everything you expect.  Excellent effort.  If you turn that into taking small steps and testing them, I think you will do well.

starter/script.js
mattscilipoti, line 8: Compare this prompt and conditional to the first prompt and corresponding conditional.  What is different?

PR #148, opened 2016-02-18
adventure homework

ghost:
complete 2
comfort 2

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,33 @@
-// put code here!
+ alert("Welcome My name is Sarah")
+  var name prompt("What is your name?")
+  alert("hello"+ "name")
+  prompt("would you like to begin an adventure")
+  var intial answer(y,n){
+    if(answer == "y");
+    var response prompt("let begin");
+    else(answer == "n")
+    var response prompt("It will be fun, you don't want to miss out!!");
+  }
+var begin prompt("Pick a charater, Katherine or James")
+var Katherine("k")
+var James("j")
+if(begin == "k");
+var response prompt("Great choice");
+else( answer =="j");
+var response prompt("Let's Begin");
+
+Alert("While trying to cook, the apartment suddenly brust in to flames")
+var action prompt("We should try to put the fire out. What should we do run,put out fire"
+ )
+ var run("r")
+ var put out fire("p")
+ if (action == "r");
+ var response prompt("you forgot you left you kitten, you have to go back");
+ if (action =="p");
+ var reponse alert(" you discovered that you did start the fire after all. The fire was started by miniture fire breathing golbins. ");
+ if else("p" == alert;)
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,33 @@
-// put code here!
+ alert("Welcome My name is Sarah")
+  var name prompt("What is your name?")
```
RobertAKARobin, line 2: Watch your indentation! Things should only be indented to the right if they're "inside" something else. For example:

```
function yo(){
  alert("I'm indented because I'm inside the function")
}
if(x === 3){
  alert("I'm indented because I'm inside the if statement")
}
```

starter/script.js
```diff
@@ -1 +1,33 @@
...
+  alert("hello"+ "name")
+  prompt("would you like to begin an adventure")
+  var intial answer(y,n){
```
RobertAKARobin, line 5: Declaring variables will ALWAYS look like this:

```
var variableName = something;
var variableName = resultOfAFunction();
```

starter/script.js
```diff
@@ -1 +1,33 @@
...
+    if(answer == "y");
+    var response prompt("let begin");
+    else(answer == "n")
```
RobertAKARobin, line 8: You need curlies `{` after `if` and `else`:

```
if(answer == "y"){
  var response = prompt("let begin");
}
```

starter/script.js
```diff
@@ -1 +1,33 @@
...
+    var response prompt("It will be fun, you don't want to miss out!!");
+  }
+var begin prompt("Pick a charater, Katherine or James")
+var Katherine("k")
+var James("j")
+if(begin == "k");
+var response prompt("Great choice");
+else( answer =="j");
```
RobertAKARobin, line 16: `else` never has anything in parentheses after it -- only `if` and `else if`.

```
if(begin == "k"){
  var response = prompt();
}else if(answer == "j"){
  var response = prompt();
}else{
  // This is a "catch-all" for when none of the other conditions are met
}
```

starter/script.js
```diff
@@ -1 +1,33 @@
...
+ if (action == "r");
+ var response prompt("you forgot you left you kitten, you have to go back");
+ if (action =="p");
+ var reponse alert(" you discovered that you did start the fire after all. The fire was started by miniture fire breathing golbins. ");
+ if else("p" == alert;)
+ var response prompt("What do you want to do  run or fight");
+ var fight("f");
+ if else ("r"=="p");
+
+ Alert(" The great battleadventure")
```
RobertAKARobin, line 33: `alert` is lowercase. Javascript is very case-sensitive.

RobertAKARobin:
Good shot, Sarah! Remember: If your code doesn't work, stop and fix it. Don't go on until it works.

PR #150, opened 2016-03-19
marygriffus_solution

marygriffus:
comfort: 5
completeness: 5

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,176 @@
-// put code here!
+// one hipster's tale
+
+// functions:
+function getAdventure(adventureState) {
+//   input: adventureState, player info
+//   effect: display current adventure
+  console.log(adventureText[adventureState]);
+};
+
+function getResponse(adventureState) {
+//  input: question/response key, playerinfo
+//  returns: option letter
+  getAdventure(adventureState);
+  var a = '';
+  console.log(adventureQ[adventureState]);
+  while (a != 'a' && a != 'b' && a != 'c') {
+    a = prompt(adventureChoices[adventureState]);
+  };
+  if (a === 'a') {
+    player['hipsterIndex'] += 3;
+  }
+  return a;
+};
+
+// game text/info:
+var playerTypes = {
+  'a': 'Foodie',
+  'b': 'Hipster',
+  'c': 'Crafter',
+  'd': 'Normcore'
+};
+
+var adventureText = {
+  'greeting': '"Hi there! I\'m Jans. Welcome to square."\nA small man with elaborate facial hair and a plaid shirt stands before you.\nLooks like you\'re new around here. This is a pop-up art space incubator slash maker innovation lab slash artisan food market."\nPeople mill about around you, sipping beers and looking at tasteful home goods.',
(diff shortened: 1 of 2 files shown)


starter/script.js
```diff
@@ -1 +1,176 @@
-// put code here!
+// one hipster's tale
+
+// functions:
+function getAdventure(adventureState) {
+//   input: adventureState, player info
+//   effect: display current adventure
+  console.log(adventureText[adventureState]);
```
tylercrosse, line 7: This is a good use case for the bracket accessor.

starter/script.js
```diff
@@ -1 +1,176 @@
...
+//  input: question/response key, playerinfo
+//  returns: option letter
+  getAdventure(adventureState);
+  var a = '';
+  console.log(adventureQ[adventureState]);
+  while (a != 'a' && a != 'b' && a != 'c') {
+    a = prompt(adventureChoices[adventureState]);
+  };
+  if (a === 'a') {
+    player['hipsterIndex'] += 3;
```
tylercrosse, line 20: The dot accessor is typically cleaner and easier to read unless you are using a variable.

``` js
player.hipsterIndex += 3;
```

Also see [this stack overflow article](http://stackoverflow.com/questions/4968406/javascript-property-access-dot-notation-vs-brackets#answer-4968448)

starter/script.js
```diff
@@ -1 +1,176 @@
...
+//   input: adventureState, player info
+//   effect: display current adventure
+  console.log(adventureText[adventureState]);
+};
+
+function getResponse(adventureState) {
+//  input: question/response key, playerinfo
+//  returns: option letter
+  getAdventure(adventureState);
+  var a = '';
```
tylercrosse, line 14: Using a more semantic variable name like `answer` or `response` would help the readability, particularly for things like `a != 'a'`

starter/script.js
```diff
@@ -1 +1,176 @@
...
+var playerTypes = {
+  'a': 'Foodie',
+  'b': 'Hipster',
+  'c': 'Crafter',
+  'd': 'Normcore'
+};
+
+var adventureText = {
+  'greeting': '"Hi there! I\'m Jans. Welcome to square."\nA small man with elaborate facial hair and a plaid shirt stands before you.\nLooks like you\'re new around here. This is a pop-up art space incubator slash maker innovation lab slash artisan food market."\nPeople mill about around you, sipping beers and looking at tasteful home goods.',
+  '1': 'It looks like there are lots of things to see around here. Some sort of bicycle race/Instagram contest is taking place in the outdoor plaza before you. Live music is drifting from a small stage. There are also some food vendors set up in tents.',
```
tylercrosse, line 35: I really like how you are able to access properties in these 'Text, Question, & Choices' objects based on the current `adventureState`. Another nit-picky refactoring opportunity is to combine them all into a single object. I found myself jumping back and forth a bunch between objects in reading this code. I think it might help the readability to have all the data for a specific state in the same location, doing something like this:

``` js
var adventure = {
  '1a':  {
    text: 'some text',
    q: 'some question',
    choice: 'some choice'
  }
}
```

tylercrosse:
You went above and beyond on this assignment and that's awesome to see! I left a couple small suggestions but I had to really look for stuff. Great job :+1:

PR #153, opened 2016-03-21
zaneh_solution

zanehmw:
comfort_level: 3.5
completedness: 4ish

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,10 +1,10 @@
 name  = prompt("Hey, what's your name?")
-age = prompt("How old are you?") 
+age = prompt("How old are you?")
 
 
 
 for(var i = age; i <= 17; i++) {
-name = "Young person";
+name = "Young person (you have played the lottery illegally)";
 }
 
 
--- a/script.js
+++ b/script.js
@@ -1,0 +1,38 @@
+name  = prompt("Hey, what's your name?")
+age = prompt("How old are you?") //I'm just going to assume everyone who plays this game is over 18
+
+//I can't figure out why this loop doesn't work properly
+// while (age < 18) {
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -0,0 +1,36 @@
...
+        if (paidOffDebts == '1') {
+          console.log ("This cycle never ended and people kept asking you for money. You eventually gave away all of your 2 billion dollar fortune. Unfortunately, you are now struggling to survive.")
+        }
+        else if (paidOffDebts == '2') {
+          console.log("The security team you hired paid off and no one has threatened you or tried to harm you since. You also hired a great lawyer who has your back no matter what. You are still very rich and occassionally go on a shopping spree or two but you have kept a majority of your wealth. You are happy!")
+        }
+    }
+
+
+else if (entryPath == "2") {
```
tylercrosse, line 26: This block (lines 26-35) should be indented another level to be at the same level as the other 'entryPath' condition on line 15

script.js
```diff
@@ -0,0 +1,36 @@
+name  = prompt("Hey, what's your name?")
+age = prompt("How old are you?")
+
+
+
+for(var i = age; i <= 17; i++) {
+name = "Young person (you have played the lottery illegally)";
+}
```
tylercrosse, line 8: I love that you used a loop, unfortunately this loop isn't really doing much. An example of something potentially more useful:

``` js
while (age < 17) {
  alert("You are not old enough, come back when you are older")
  age = prompt("How old are you?")
}
```

This will loop until the user enters an age that is 18 or older

tylercrosse:
Nice work :+1: this was a fun one to grade!
Feel free to ask questions or add comments after you list comfort_level & completeness. Is there anything that would help you feel a little more comfortable?

PR #155, opened 2016-03-21
Lauren's solution

lknapalolo:
comfort level: 4
completeness: 5


starter/script.js
```diff
@@ -1 +1,74 @@
...
+    //Question 3b: Workout Plans
+    else if (fitOrFood == "W" || fitOrFood == "w"){
+      //Question 3b; HIIT Track
+      var workout = parseInt(prompt("Choose your workout style:\nPlease type the number that corresponds with your answer.\n\[1\] HIIT  \n\[2\] Yoga"));
+        if(Number(workout) === 1){
+          var hiitNameArray = userName.split("");
+          var i=0;
+          document.write("<p  class=\"title\">" + "Hey, "+ userName + "!" + "<p class=\"subtitle\">" + "Here's your customized workout, based on the letters in your name." + "<p>");
+              while(i < hiitNameArray.length) {
+                  if(hiitNameArray[i] === "A" || hiitNameArray[i] === "a"){document.write("50 jumping jacks")}
```
andrewsunglaekim, line 32: What are some ways we can start DRY up this code? One way is to lower case `hiitNameArray[i]`. Then all we have to do is write a condition for the lower cased version. 

Additionally we recently learned about objects, how can we use an object instead of all these if/else conditionals?

starter/stylesheet.css
```diff
@@ -0,0 +1,44 @@
+html {
```
andrewsunglaekim, line 1: styles!? nice work

andrewsunglaekim:
Great job :+1:

PR #156, opened 2016-03-21
WDI 9 HW Week 2 Day 1

lelandjordon:
Completeness: 4
Comfort: 2
Pardon the not-exactly-socially-progressive questions in the story, but I wanted to keep it somewhat simple.

### DIFF
--- a/cyoa_story.txt
+++ b/cyoa_story.txt
@@ -1,0 +1,35 @@
+What's your name?
+* 'Bill'
+
+Are you male or female? **(M/F)**
+(F)emale:  Do what you want to do today!
+
+(M)ale: Are you married? **(Y/N)**
+* 'No'
+  Do you have a girlfriend? **(Y/N)**
+    * 'N'
+      * 'N' >-
+          You're on your own today.  Movies, video games, or
+          food?  Or all three? Or are you broke? **(M,V,F,3,B)**
+          (M)ovie: Great!  Go to the cheap theater and spend $15 on
+                  popcorn for dinner.
+          (V)ideo games:  Great!  Stay home, play games, and spend
+                  $15 on pizza for dinner.
+          (F)ood:  Great!  Order pizza delivery for $15.
+          (3):  Great!  Dave & Buster's it is!  ...If you have $115 to blow. **(Y,N)**
+            (Y)es: Party!
+            (N)o:  Stay home, play video games, and buy pizza AND beer!
+          (B)roke: Stay home, play video games, and download a movie while you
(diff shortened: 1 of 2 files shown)


cyoa_story.txt
```diff
@@ -0,0 +1,35 @@
...
+*(Yes):  Ask what she wants to do.  Shopping, movie, dinner, or stay home?
+        **(S,M,D,H)**
+        (S)hopping: Cool!  Do you want to go to the mall, or to Costco?  **(M,C)**
+            (M)all:  Go to Tyson's Galleria.
+            (C)ostco:  Go to Costco, get some toilet paper, then
+                      continue to Tyson's Galleria.
+        (M)ovie: Cool!  Do you want to see Fast & Furious 12 or The Notebook?
+        **(F/N)**
+          (F)ast & Furious: Really?  Awesome!
+          The (N)otebook: You get to show your sensitive side. Bring tissues!
```
andrewsunglaekim, line 35: :+1: on the pseudo code, i think you're the only one that i've seen taht wrote some! great work

starter/script.js
```diff
@@ -1 +1,125 @@
 // put code here!
+var name = prompt("What\'s your name?");
+console.log(name);
+
+var gender = prompt("Hi, " + name + ". You have the day off today!  This script is going to help decide what to do.  First, we have to ask: are you male or female? (M/F)");
+//console.log(gender);
+ //**(M/F)**
+
+if (gender === 'F' || gender === 'f'){
```
andrewsunglaekim, line 9: how can we limit this to just `(gender == "f")`? if we lowercase the users input, we only have to have one portion of the conditional.

andrewsunglaekim:
Good work leland! One thing I will note is that you have both a console and alert to message the user, I think you just need one or the other. I noticed that you put a comfort level of 2 though. Your hw is evidence of a stronger comfort level, but let me know what things you are still uncomfortable with and perhaps i can shed some light. Anyways, great work! :+1:

starter/script.js
lelandjordon, line 9: There's a function to automatically lowercase the input?

lelandjordon:
It took several instances of stepping away from the computer and letting my mind flush out all the symbols before I could piece some of the things together.  The two breakthroughs were 1) learning that an evaluator like (a == b || c) will never work, and 2) learning how to "turn off" big chunks of code by commenting it out with /\* */.  I would have never realized I was missing a } halfway through if I hadn't turned off the rest of the script.

PR #157, opened 2016-03-21
final commit

7ruth:
Completeness: 4
Comfort: 3


andrewsunglaekim:
first off, this was really cool! Awesome job trying out google maps.

One thing that I wanted to mention, though, was that it fell short of the requirements of this assignment.  Thats ok though! The coolness factor makes up for it.

Couple of things that I do want to make sure:
- There are a lot of dependencies that were added to this program, do you have a solid understanding of how all these things are integrating with each other?
- Seems like a whole lot of code for the sparkle effect, was this hand rolled?(either way is cool, much cooler if handrolled!)

Anyways really great job! Going forward, our exercises we'll be come a bit further scoped, be sure to try and follow the parameters as well as possible. :+1:

7ruth:
Thanks for taking a look Andy, 

I knew that it wasn't up to all requirements and that was due to me creeping scope so much...I bit too much and start to slowly realize that as time was approaching 10 pm on Sunday...you did say to submit whatever we had and I did that when I felt like it was getting close. 

I adjusted the positioning for sparkle effect but took structure from a codepen I saw a while back. 

I do feel pretty comfortable with the dependencies used in this code to make sure all js and google work properly.

PR #158, opened 2016-03-21
bare bones of morning to go

habbywhyte:
completeness: 2
comfort: 2.5

Still making changes.  Not ready for js validator yet.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,33 @@
 // put code here!
+var tantrumFree = function() {
+  console.log("**************************")
+}
+console.log ("You've overslept one morning and realized you have 25 minutes to get yourself and your kindergartener out the door")
+var wakeUpInput = prompt("You can either jolt out of bed or push the snooze button") (Please enter 'j' or 's')
+
+if (wakeUpInput) == "j"{
+  console.log("You jolt out of bed and begin dressing your kindergartener, when she decides she doesn't want to wear your option but rather her polka-dot pink Doc McStuffin shirt with her green striped duck pants."
+}
+var (reponseToToddler) = prompt("You realize that this very moment could make or break her blossoming kindergarten reputation" and decide to \n('a') - argue with her. We're talking reputation here people!...hers and yours!\n('g') - give in, toddlers have no reputation.  This may actually increase her self-confidence, while the other parents question your parenting skills\n('n')- negotiate with her.  She may know the latest trends that you may not be hip to and needs just a little guidance.")
+
+if(responseToToddler == 'a'){
+  console.log("You argue tooth and nail with her, when you realize you've wasted time, increased your blood pressure and will inevitably be late")
+  buffer()
+}
+
+else if (responseToToddler == 'g'){
+  console.log("With extra vigor, she happily puts on her clothes, saving you the extra minutes to brush your teeth.  Who cares what she wears anyways, you say to yourself")
+  buffer()
+  var toddlerReaction == prompt ("She asks you what's for breakfast?"\n('p') - Pancakes\n('o')'Not Pancakes');
+  if (responseToToddler =='p') {
+    console.log("She gives you her toothiest grin and runs down the stairs.  You follow her down the stairs, to pop in her pancakes in the oven, saving you a minute to finish getting dressed.");
+    tantrumFree();
+  }
+else if (responseToToddler =='o'){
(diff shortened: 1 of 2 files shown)


starter/script.js
```diff
@@ -1 +1,33 @@
...
+var tantrumFree = function() {
+  console.log("**************************")
+}
+console.log ("You've overslept one morning and realized you have 25 minutes to get yourself and your kindergartener out the door")
+var wakeUpInput = prompt("You can either jolt out of bed or push the snooze button") (Please enter 'j' or 's')
+
+if (wakeUpInput) == "j"{
+  console.log("You jolt out of bed and begin dressing your kindergartener, when she decides she doesn't want to wear your option but rather her polka-dot pink Doc McStuffin shirt with her green striped duck pants."
+}
+var (reponseToToddler) = prompt("You realize that this very moment could make or break her blossoming kindergarten reputation" and decide to \n('a') - argue with her. We're talking reputation here people!...hers and yours!\n('g') - give in, toddlers have no reputation.  This may actually increase her self-confidence, while the other parents question your parenting skills\n('n')- negotiate with her.  She may know the latest trends that you may not be hip to and needs just a little guidance.")
```
adambray, line 11: Note that this line has a syntax error, that's because you've ended the quote marks around the string after "blossoming kindergarten reputation". Be cause of this, JS is trying to treat the following words "and decide to ..." as _code_ rather than text in your program. This would almost certainly throw off the rest of the program.

starter/script.js
```diff
@@ -1 +1,33 @@
...
+
+if(responseToToddler == 'a'){
+  console.log("You argue tooth and nail with her, when you realize you've wasted time, increased your blood pressure and will inevitably be late")
+  buffer()
```
adambray, line 15: I don't see a `buffer` function defined anywhere in your code... was that something you made and then deleted?

starter/script.js
```diff
@@ -1 +1,33 @@
...
+  console.log("With extra vigor, she happily puts on her clothes, saving you the extra minutes to brush your teeth.  Who cares what she wears anyways, you say to yourself")
+  buffer()
+  var toddlerReaction == prompt ("She asks you what's for breakfast?"\n('p') - Pancakes\n('o')'Not Pancakes');
+  if (responseToToddler =='p') {
+    console.log("She gives you her toothiest grin and runs down the stairs.  You follow her down the stairs, to pop in her pancakes in the oven, saving you a minute to finish getting dressed.");
+    tantrumFree();
+  }
+else if (responseToToddler =='o'){
+  console.log("Your toddler lets out a loud shriek and begins to whine.  She then refuses to take a bite, making you late.");
+  bumper();
```
adambray, line 28: (Same as what I wrote above re: `buffer`)

starter/script.js
```diff
@@ -1 +1,33 @@
 // put code here!
+var tantrumFree = function() {
+  console.log("**************************")
+}
+console.log ("You've overslept one morning and realized you have 25 minutes to get yourself and your kindergartener out the door")
+var wakeUpInput = prompt("You can either jolt out of bed or push the snooze button") (Please enter 'j' or 's')
```
adambray, line 6: Similar to the comment I made a few lines below, this line has issues with quotations. The key thing to realize is that `prompt` is a function that takes one argument (a string). This means when we call it, we have to give it _all_ the text we want it to show in 1 string:

``` js
//incorrect:
// var wakeUpInput = prompt("You can either jolt out of bed or push the snooze button") (Please enter 'j' or 's')

//correct:
   var wakeUpInput = prompt("You can either jolt out of bed or push the snooze button (Please enter 'j' or 's')" )
```

Even though our string of text has parenthesis in them, it doesn't mean we need to end the string, they are _part of the string_, and distinct from the closing parenthesis that we use to _call_ the function `prompt`.

adambray, line 6: Note to that if we want to use quotes _inside our string_, we have to use single quotes, because double quotes would _end_ our string rather than be part of it.

(You got this correct, but I wanted to call it to your attention).

adambray:
Hi Habby, this is a pretty good start, but definitely has some issues, that seem to be causing the program not to run.

In general, I would still suggest:
1. Take _smallsmallsmall_ steps. Add only one thing at a time (usually this would involve adding or editing no more than 3 lines of code) and then _test_ and don't do anything else until you get that little part of the program working.
2. After each step or two, commit. That way, you can always go back, and it helps me see your process (I can look at the history) and see where you started having more trouble.

PR #159, opened 2016-03-21
Changes made to homework for ChooseYourOwnAdventure.

mvaden:
Completeness: 2
Comfort: 2.5

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -31,7 +31,9 @@ if year.to_i >= 2015
 
       elsif action == "r"
          puts "`R`un like a coward. You get away, but your future son Marty Jr. is heckled for the rest of his days for his dad's cowardice."
+
       end # s/r
+
 
    elsif name == "g"
       puts "Griff is asking you if you are in, or out. What do you say? **(I/O)**"
@@ -43,6 +45,7 @@ if year.to_i >= 2015
 
       elsif asking == "o"
          puts "`O`ut. Good call. You deck Griff in the jaw and run away. He gives chase on his hoverboard and ends up in a pile of manure."
+
       end # i/o
 
    end # b/g
@@ -70,12 +73,15 @@ elsif year.to_i >= 1955 && year.to_i <= 1984
       puts "Interesting. You set up an elaborate plan for your future Dad to surprise your Mom by beating you up. Despite going horribly awry, the plan ultimately works. You may go back to your own time."
 
    end # y n s
+
 end # 1955 - 1984
+
 
 elsif year.to_i < 1955
    puts "I see you're a fan of Back to the Future 3. You've run out of nitroglycerin to get back to your own time. How do you power the Time Machine? **(H/M/T)."
 
-   power = gets.chomp
+
+   power == gets.chomp
    puts power
 
    if power == "h"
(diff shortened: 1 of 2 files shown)


tylercrosse:
I love seeing the psuedocode! You are definitely on the right track with assignment. I really encourage you to continue to work on this a little more and make another commit. I'll leave the pull request open for the time being. Please let me know if you're having trouble with a particular part of it!

mvaden:
Update: Choose Your Own Adventure In ruby.
Thanks for your help, Tyler!
Only problem was a syntax error on line 80, col 1. No other noted errors.
Comfort Level: 4/5
Completed-ness: 4.8/5

app.rb
```diff
@@ -0,0 +1,108 @@
...
+
+   elsif dance == "n"
+      puts "Honorable. But this also means that your future Dad will never meet your Mom, and therefore you cannot exist."
+
+   elsif dance == "s"
+      puts "Interesting. You set up an elaborate plan for your future Dad to surprise your Mom by beating you up. Despite going horribly awry, the plan ultimately works. You may go back to your own time."
+
+   end # y n s
+
+end # 1955 - 1984
```
tylercrosse, line 77: Line77 is the source of your error. We can't start an `elsif` without an if statement before it. Here are some valid formats for conditionals, notice they all only say `end` once and comes at the end (named that for a reason)

``` rb
#example 1
if (condition)
  # do stuff
end

#example 2
if (condition)
  # do stuff
else
  # do other stuff
end

#example 3
if (condition)
  # do stuff
elsif (another condtion)
  # do other stuff
end

#example 4
if (condition)
  # do stuff
elsif (another condtion)
  # do other stuff
else
  # some other other stuff
end
```

app.rb
```diff
@@ -0,0 +1,108 @@
...
+      moment == gets.chomp
+      puts moment
+
+      if moment == "t"
+         puts "Interesting choice. Unfortunately the Doc can't grab Clara and get back to the car in time. He ends up staying in 1855 with her."
+
+      elsif moment == "l"
+         puts "Smart choice. Unfortunately the Doc was deeply in love with Clara, and when he gets back to 1985 he becomes very depressed."
+
+   end # t l
```
tylercrosse, line 106: This end should be indented over to match the if statement that started this block

app.rb
```diff
@@ -0,0 +1,108 @@
...
+
+end # h m t
```
tylercrosse, line 108: This one should also be indented over to match the if statement for `h`. You also then need to include an end statement to finish the block of conditionals for years.

app.rb
```diff
@@ -0,0 +1,108 @@
...
+   end # y n s
+
+end # 1955 - 1984
+
+
+elsif year.to_i < 1955
+   puts "I see you're a fan of Back to the Future 3. You've run out of nitroglycerin to get back to your own time. How do you power the Time Machine? **(H/M/T)."
+
+
+   power == gets.chomp
```
tylercrosse: This won't work. Care to guess why?

tylercrosse, line 84: This won't work. Care to guess why?

app.rb
```diff
@@ -0,0 +1,108 @@
...
+   if power == "h"
+      puts "Good idea, but no. The time machine needs to get to 88mph. 12 horsepower ain't gonna cut it."
+
+   elsif power == "m"
+      puts "You'd be better off drinking the moonshine. Do not pass Go, do not collect $200. Stuck in 1855."
+
+   elsif power == "t"
+      puts "Good call! This plan seems to be working. But wait! Clara wants to go Back to the Future with you at the last moment. What do you do? **(T/L)**"
+
+      moment == gets.chomp
```
tylercrosse, line 97: Same issue as line 84

tylercrosse:
Nice work! :+1:  You need to really pay attention to what the error messages are telling you and resolve them before you continue to write other code. Otherwise you have no way of knowing if any additional code you write works or not. I would love for you to go back and fix all of these issues over the weekend and resubmit it on Monday. You should be able to get this to work all the way through.

PR #160, opened 2016-03-21
Javascript and HTML

JoshNezi:
I have a few questions about the syntax. Some prompts come up and others don't.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,0 +1,50 @@
+//name request//
+var person = prompt("Please enter your name", "Kevin Hart");
+
+if (person != null) {
+    document.getElementById("demo").innerHTML =
+    "Hey! " + person + "! Are you ready for a ride along with Ice Cube?";
+}
+
+var text;
+var favSong = prompt("What's your favorite Ice Cube Song");
+switch(favDrink) {
+    case "It was a good day":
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -0,0 +1,50 @@
...
+var person = prompt("Please enter your name", "Kevin Hart");
+
+if (person != null) {
+    console.log("Hey! " + person + "! Are you ready for a ride along with Ice Cube?");
+
+}
+
+var text;
+var favSong = prompt("What's your favorite Ice Cube Song");
+switch(favSong) {
```
andrewsunglaekim, line 11: One thing i would mention, it I would make the prompt something a little bit more manageble for users. Give them just one letter to enter that represents the song. Long strings like the one in your case statements are much more error prone

andrewsunglaekim:
As your code executes serially( in order ) Nothing executes based on the result of a prior question. I encourage you to look at the solution to see how you could redo this assignment to have more user involvement with the story line. When code doesn't execute as we intended, it's up to you as the developer to try and figure out why and attempt a solution. I encourage you to look into the solution branch of this repo and redo this exercise at some point to really hammer down these JS fundamentals.

Good work trying out those switch statements! :+1:

PR #161, opened 2016-03-21
First draft of adventure program

LOHankins:


starter/script.js
```diff
@@ -1 +1,168 @@
-// put code here!
+// My adventure program. Select your vacation destination
+// based on my travels!
+
+var news = prompt("Its 7am.  Time to wake up to another day of city life.  You are dressed and ready to go to work.  Do you watch the morning news before leaving?");
+var email = prompt("Do you check your email before leaving");
+var dog = prompt("do you walk, feed and play with your dog before leaving?");
+var dead = 'no';
+var destination = null;
+var info = {"news":"no", "email":"no", "dog":"no", "dead":"no"}
```
andrewsunglaekim, line 9: almost seems like a duplication of your `newsEmail` function

starter/script.js
```diff
@@ -1 +1,168 @@
...
+		email = prompt("do you check your email before doing anything else");
+		text = prompt("do you check your text messages before moving on");
+	}
+
+}
+
+
+console.log("almost end of program");
+
+for (var i = 7; (i < 24); i += 5) {
```
andrewsunglaekim, line 122: although `i` is a convention we follow, i think that there my be a better variable for the purpose of this code base, maybe hour or something.

andrewsunglaekim:
I added a couple of comments that I noticed, but i have some overarching comments and suggestions. 

It's a really cool concept of leveraging the hour of the day. And I love that you're leveraging functions to abstract your functionality.

Think smaller at first, get things to work exactly as you want, than grow your application from there. 

Other than that, great work!

PR #170, opened 2016-06-01
Gandy Choose Your Own Advernture HW

gandyland:
confort: 3
completeness: 3


RobertAKARobin:
Definitely on the right track, but you've got some inconsistent indentation. This is a result of and/or causing you to miss the places where you need closing curly braces. Make use of Atom's auto-indent function to help.

✅

PR #171, opened 2016-06-02
Made a Judge Dredd text game

chapel-of-bools:
Completeness: 4 / 5
Comfort Level: 3 / 5

I attempted to add more while loops but ended up with infinite loops which broke the game.  Any advice on how to make a loop refer back to a previous input variable without this happening?


RobertAKARobin:
Hmm. I'm not entirely sure what you mean. If you can give me some example code I'd be happy to look!

✅

PR #172, opened 2016-06-02
Welcome to Westeros!

AT89:
Not really finished but I hit my timebox. I wanted to add images for each function to help guide the choices. 
As of now, the code won't run due to syntax errors but I will look into it tomorrow.

4/5 comfort level - got familiar with writing more functions and var's
4/5 completeness - i couldnt get all the choices in that I wanted, I wanted to add another family story and more destinations.

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -1,10 +1,15 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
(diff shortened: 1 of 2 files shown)


RobertAKARobin:
✅ Next time, fix syntax errors as you go. **Never ever ever ever ever write more than 2 lines of code without stopping to make sure it still works.**

PR #174, opened 2016-06-02
Julia Najafi Adventure 

JuliaNajafi:
comfort level: 4/5
Completeness: 4/5

I wanted to prompt my user to write in a number, but couldn't get it to work in my code. I wanted my If statement to be If (variable === number which corresponds to choice), but it wouldn't work after trying different things. So I stuck with using "a", "b" and "c". How can I incorporate numbers into my choices using parseInt()? 
Thanks!

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>Document</title>
+  <title>Zombie Apocalypse</title>
 </head>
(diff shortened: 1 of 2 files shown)


RobertAKARobin:
``` js
var userInput = parseInt(prompt("Answer this question!"));
```

Or:

``` js
var userInput = prompt("Answer this question!");
userInput = parseInt(userInput);
```

✅

PR #177, opened 2016-06-02
Julia Schulltz CYOA JS HW

jschultz21:
_question_ - is there anywhere where I could have used switch cases to be more efficient?

Comfort level: 4
completeness: 5


RobertAKARobin:
Anyplace you have something like:

``` js
if(blah === "a"){
...
}else if(blah === "b"){
...
}else if(blah === "c"){
...
```

...and so on, you can use `switch`. I don't really like switch very much though because its syntax rules are really strange. But whatever floats your goat.

✅

PR #180, opened 2016-06-02
Dave's Adventure

dcordz:
Completeness = 4/5? (I think the requirements asked for a few more ifs)
Comfort = 4.5? (It still seems like magic sometimes)


RobertAKARobin:
✅ 

There will always be magic.

PR #182, opened 2016-06-02
added pseudo code showing my idea

arthuradams:
I only wrote pseudo code, but I think I came up with a cook way to do this.


RobertAKARobin:
✅

PR #186, opened 2016-07-02
Choose your own adventure - vacation

megancoyle:
Comfort score: 5
Completeness: 5

I feel like I repeated myself a lot in this, so if there's a more efficient way that I could have written this, let me know!

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -34,3 +34,14 @@ var daysVacation = parseFloat(prompt("You’ve had a long year, it’s time for 
       } else {
         alert("Sorry, but you didn't pick one of the possible locations!");
       }
+
+        // More Options in Europe
+        if (moreInEurope == "Germany") {
+          alert("Time to eat brats and drink beer!");
+        } else if (moreInEurope == "England") {
+          alert("Great, let’s book your trip! Maybe you’ll get a chance to see the Queen of England!");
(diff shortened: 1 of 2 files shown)


starter/script.js
```diff
@@ -1 +1,71 @@
...
+      var whereInEurope = prompt("Ohh, Europe! how about going to France or Italy?");
+      // Options for if user wants to go to Europe
+      if (whereInEurope == "France") {
+        alert("Oiu! Let’s get you set up with going to the Eiffel Tower and make some fine dining dinner reservations!");
+      } else if (whereInEurope == "Italy") {
+        alert("Time to get ready to see some fine art and enjoy that great Italian cuisine!");
+      } else if (whereInEurope == "No") {
+        var moreInEurope = prompt("How about going to Germany or England?");
+          // More Options in Europe
+          if (moreInEurope == "Germany") {
```
andrewsunglaekim, line 27: should keep this and following conditions on the same tab length as the var statement in line 25

andrewsunglaekim:
yeah, unfortunately that's kind of the nature of this assignment. You could argue using switch statements would be more efficient. You could also argue encapsulating some of this logic in functions or even an object. But at the end of the day, the assignment was just to make sure students have a solid understanding of control flow. Which you clearly do! Great work. 👍

PR #190, opened 2016-07-04
Tyler Rosenberg Choose Your Own Adventure

TyRyanR:
Comfort: 4.5
Completeness: 4.5

Had a fun time with this one. Got a little confused with methods that allow you to slice individual characters out of an array.


nayana487:
Nice job Tyler! Like how you nested your conditionals here--that can sometimes trip people up, but you did a good job here. Your code was also really well formatted and easy to read. Did you have a specific question about using the slice() method? Typically, you can pass one index number as a parameter to return that value or two with the first index being where you start and the second indicating where it would end. Important note--it doesn't actually include the last index specified, just up to that point. Good demo of that here: http://www.w3schools.com/jsref/jsref_slice_array.asp

PR #191, opened 2016-07-04
Chris Phillips Political Adventure!

ghost:
Comfort: 4
Completeness: 4.0

Alright so I didn't really like using the prompt when i realized I was having issues loading it with pictures etc. I also wanted to use vanilla JS in the spirit of the exercise so learning the jQuery library was out for this one.

I googled a bit and found a few ways people had incorporated a full text adventure experience into websites and put this together. Basically this structure allows for a bunch of scenarios to be set in an object. This object is accessed by a function that changes our core div elements that correspond with the different values in each scenario (text, image, buttons, input).

The best part about this was learning how an onkeypress function worked and customizing it to work with the input boxes (for when we wanted to make use of both string and number user inputs). I think I'm going to come back to this before tomorrow so I can customize the onkeypress function a bit more to be more interesting. Right now it's just an if elseif statement that checks for either a number or text. For the text I want to do an RNG function that will pick between two scenarios (we could even write this to select between a fair amount of scenes depending on how fancy we wanted to get). For the number input I want to just design an easy if else if that changes based on what number is put in. It's a lot of if else if nesting but would be very cool.

I'll do this tonight when I come back to this assignment. I'm going to hit up the calculator for now.

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -8,8 +8,8 @@
 <body>
     <div id="images"></div>
     <div id="text"></div>
-    <input id="input" onkeydown='submit candidates name' placeholder='Type what campaign you are joining'>
     <div id="buttonBox"></div>
+    <input id="input"></input>
 </body>
 <script src="script.js"></script>
 </html>
--- a/starter/index.html
+++ b/starter/index.html
@@ -1,9 +1,9 @@
 <!DOCTYPE html>
 <html>
 <head>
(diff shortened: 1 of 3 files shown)


nayana487:
Hey Chris, 

awesome job here! No real criticisms here. Love how you organize your code--so clean and easy to read. Also really cool how you incorporated the scenario object. Nice work also using onkeypress.

PR #193, opened 2016-07-04
Sarah Shultis - Seinfeld Adventure

skshultis:
comfort: 3.5/5
completeness: 4/5

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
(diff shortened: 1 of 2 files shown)


nayana487:
Hey Sarah, 

This was fun! really enjoyed playing your game. A thought on your code: when you're writing conditional statements, start with "if" as the first condition, "else if" for all following conditionals except the last one, and "else" --the catch all  for the leftovers--for the last one.

PR #194, opened 2016-07-05
Taylor Edwards HW (7/4)

ghost:
Completeness: 3/5
Comfort: 3/5

Comments: The hardest part of this exercise was by far the initial conceptualization, i.e. planning out the "roadmap" my code would take. For me, part of this challenge was that I wanted something creatively satisfying along with functional code - and I had to eventually let go of my original idea in favor of functional code.

Even having gotten to that point, though, I'm not at all confident that this code is as clean, DRY, or effective as it could be... and again, I feel like if I had been able to get a better top-down view of the project, I probably could have created some functions or compressed some code to make it more effective.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Adventure</title>
+</head>
+<body>
+
+  <script src="script.js"></script>
+</body>
+</html>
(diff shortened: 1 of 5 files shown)


andrewsunglaekim:
This looks great, albeit very ummm...exactly the prompt? I guess what I'm trying to get at, is we're not trying to kill you with the requirements. The idea was to have a little bit of fun with it. That said the code looks great. the only thing i can think of refactoring wise is maybe leveraging switch statements?. But really this exercise kind of lends itself to being really wet. I mean one could argue moving some of the logic into functions or into an object even. But I think this assignment was really just to make sure the student is comfortable with if/else conditionals and for loops.

PR #195, opened 2016-07-05
Evan Sanderson-  Choose Your Own Adventure

EvanSanderson:
comfort level: 5
completeness: 5


nayana487:
Hey Evan, 

Awesome job with this! Had a lot of fun playing. Just a couple small thoughts. I'd always include an else statement to act as a catch all in case someone doesn't input what you want. Also, be mindful of spacing. It looks a little odd between lines 69 to 75. I always run my code through a beautifer...like this one:http://jsbeautifier.org/.. to make sure it's readable.

PR #196, opened 2016-07-05
pushing what i've got

chaseTfreeman:
I had to set a time cap on this because I ended up wasting hours of time going nowhere. 

My comfort level was actually 3-4.. I was more concerned about the quality/flow of the "story" that I wasted a good two hours trying to figure out how to work a loop in there and in the end its all garbage. 

I'm could use some clarification on loops in this instance because it just didn't' seem natural to use a loop at all in this. 

Anyway, comfort : 3/4ish 
completeness: 2


script.js
```diff
@@ -0,0 +1,53 @@
...
+
+
+
+var ready = prompt("Wake up, Neo. Are you ready?");
+
+ if (ready != null) {
+    confirm("Great. Follow the white rabbit");
+
+  }
+  else {
```
andrewsunglaekim, line 15: make sure to indent correctly to keep this in the same tab length as the if condition it's associated with.

script.js
```diff
@@ -0,0 +1,53 @@
...
+  confirm("..but you have to be, Neo. You are the One. The Matrix has you. Follow the white rabbit. 'Knock, Knock")
+};
+
+  var meetMorpheus = prompt("After a bit of trouble, you meet Morpheus." + " " + "Morpheus says: 'Welcome, Neo. Do you know who I am?''");
+
+    if( meetMorpheus != null || "no" || "No") {
```
andrewsunglaekim, line 21: this will always evaulate to true `meetMorpheus != null || "no" || "No"` Basically what your sayins is `meetMorpheus !- null` or `"no"` or `"No"`. Because strings are truthy in javascript, this condition will ALWAYS evaluate to true.

I think what you might have wanted was  `meetMorpheus != null || meetMorpheus ==  "no" || meetMorpheus == "No"`

andrewsunglaekim, line 21: i'm not going to comment on further ones, just know that if you ever do something like this : 

``` js
someVar == 1 || 5
```

it will always return true. By virtue of the fact that the value `5` is truthy.

script.js
```diff
@@ -0,0 +1,53 @@
...
+      confirm("Great. That makes this easier.")
+    }
+
+    else {
+      confirm("In the Matrix and to the Agents I am an international criminal wanted for almost every cyber-crime in the books, if not more. However, I am not those things to you. I can be a trusted friend. I am a commander from Zion--more on that later.")
+    };
+
+var trust = prompt("Do you trust me? [type Yes or No]");
+
+   if (trust === "Yes" || "yes") {
```
andrewsunglaekim, line 31: in general, you only need to tab over if your nesting if conditions within another function block. Try not to nest unless that's the case

andrewsunglaekim:
seems like you have a good handle on the basic concepts. Definitely work on code indentation. Good work 👍

PR #198, opened 2016-07-05
Aubrey Northam's Adventure HW

aubreynortham:
Comfort: 4
Completeness: 4.8

I had to work out a few organizational challenges, but finally got all of the logic to work the way I wanted. My only "completeness" shortcoming was that I couldn't get my background image to load in the browser until after the console game is over, still not sure why. Also, I remembered to test my javascript code as I went along, but I forgot about committing that progress from the terminal, so I'll need to get better about remembering those commits.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,140 @@
-// put code here!
+/* Pseudocode in this section:
+
+intro alert:
+Well, it's been a long week. Time to WINE ABOUT IT. But wait... before you grab the nearest option, let's do some soul searching so you can find the perfect pick.
+
+first prompt:
+Let's talk lifestyle choices. Are you feeling 'fancy' or are you feeling more 'thrifty' tonight?
+fancy: Ok, well if we're being fancy, you'll need to class it up by shopping at Total Wine.
+thrifty: We can't all live the high life. Head over to Trader Joe's where you won't be judged for keeping it under $7.
+
+second prompt: Let's get the obvious out of the way: do you prefer 'red' or 'white'?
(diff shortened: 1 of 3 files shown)


starter/script.js
```diff
@@ -1 +1,140 @@
-// put code here!
+/* Pseudocode in this section:
+
```
superbuggy, line 2: AWESOME pseudocode 👍

starter/script.js
```diff
@@ -1 +1,140 @@
...
+        alert("DANGIT! You've been caught.. ditch the basket and shout- THIS ISN'T TACO BELL!! WHERE AM I?!? -and make a run for the door. Better luck next year, kid.");
+      }
+
+    }
+    //ENDPATH3
+    else {
+      alert("Welp. I don't see any twist-tops here. Looks like your night just ended.");
+    }
+  }
+}
```
superbuggy, line 140: This was super funny, especially the charity gala joke and the taco bell joke; those were my top faves and I really enjoyed playing this! Your code is well organized and commented helpfully, and cross references your pseudo. That's awesome. Being well organized makes it easier to get more done when you code.

As such I don't have any suggestions to improve your code in its current form. However, a way to rewrite it might involve storing all the strings that are used for alerts and prompts in an object full of really long strings, and then accessing properties of that object depending on the path taken by the user. I can show you what that might look like, at your request. Great work!

superbuggy:
Also, the background image loads after your game due to some side effects of alert and prompt. They will actually delay the execution of other code or HTML loading.

PR #201, opened 2016-07-05
The Wire Adventure Game_HW_07/05

ahmedelk91:
Comfort: 2-3

Completeness: 3

Comments: I had a lot of trouble with this exercise. However, I will say it was fascinating trying to relate my favorite show to a set of commands. I had a hard time finding out how to implement loops and even more trouble getting my code to work. Multiple if/else statements were also troublesome. Overall, however, I would say this exercise was great in getting me more comfortable with working with Javascript. Would love any feedback you can provide. It took me about 5 hours to complete what I have submitted so I time boxed myself. Thanks!

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,34 @@
-// put code here!
+confirm("This game contains strong language and some adult situations. Player discretion is advised...")
+
+confirm("Welcome to West Baltimore");
+
+var response = prompt("Wanna be in the game shorty? Yes or No?");
+
+if (response === "No") {
+  alert("You can't lose if you don't play. [Technical, but less fun, victory]");
+  //how do you fully stop code from running at this point?
+} else {
+  alert("Ight young hoppa, I'll put you in charge of the count.");
+};
+
+var count = prompt("It's the end of the day shorty. Is the COUNT RIGHT or did you FUCK UP THE COUNT?");
+
+if (count === "COUNT RIGHT") {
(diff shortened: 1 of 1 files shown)


superbuggy:
Good effort, although you would want to nest if statements (meaning putting if statements inside if statements). The answer to the first question doesn't actually alter the course of events, just the response in relation to the input. Also I would turn like 23 into a prompt.

starter/script.js
```diff
@@ -1 +1,34 @@
...
+var response = prompt("Wanna be in the game shorty? Yes or No?");
+
+if (response === "No") {
+  alert("You can't lose if you don't play. [Technical, but less fun, victory]");
+  //how do you fully stop code from running at this point?
+} else {
+  alert("Ight young hoppa, I'll put you in charge of the count.");
+};
+
+var count = prompt("It's the end of the day shorty. Is the COUNT RIGHT or did you FUCK UP THE COUNT?");
```
superbuggy, line 14: you could move line 14 - 29 inside of the else statement of the condition above. why would you want to to do that? one reason is that if you answered no to the first question, the program would no longer ask you any questions. that's how nested conditionals can be used. what are nested conditionals? if/else/else if statements inside of other if/else/else if statements

starter/script.js
```diff
@@ -1 +1,34 @@
-// put code here!
+confirm("This game contains strong language and some adult situations. Player discretion is advised...")
```
superbuggy, line 1: Pseudocode can help us plan our programs; that should go here according to the assignment

PR #202, opened 2016-07-05
Initial HW commit. HAving trouble inplementing all of my ideas concer…

mrnelley:
…ning game play but I do understand the principle. Cant wait to learn how to create a more complete application
 I think I did this correctly... Haha
Comprehension: 4..5
Completeness: 4

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,74 @@
 // put code here!
+var adventure = prompt("What is your name, soldier?");
+console.log(adventure);
+var Unit = prompt("what is your rank \n Please select from the following menu: \n
+\t enter E1 for Airman-Basic \n \t enter E2 for Airman \n \t enter E3 for A1C \n \t Enter E4 for NCO");
+console.log(Unit);
+var count = 0;
+var bodyCount=0;
+function myFunction()
+{
+  var x = document.createElement("INPUT");
+  x.setAttribute("type", "number");
+  x.setAttribute("value", "12345");
+  document.body.appendChild(x);
+}
+// function shoot()
+// {
+//   var add = 0;
+//   add ++;
+// }
+if (Unit == 'E1') {
+    alert('Fresh meat huh? Well you picked a hell of a time to enlist. Dooms day has begun.');
+
+  }
+else if (Unit == 'E2') {
+  alert(
+    ('Your new to this... Just out of bmt. Hahaha go home. Youre not ready for this!'));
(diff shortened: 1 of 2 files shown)


superbuggy:
`if (calc == 5 || 4 || 3 || 2 || 1)` should be `if (calc == 5 || calc ==  4 || calc ==  3 || calc ==  2 || calc ==  1)`; non-zero numbers are truthy in javascript. (`0==true` will evaluate to `false`) `if (calc == 5 || 4 || 3 || 2 || 1)` is like saying "if calc equals equals, or if 4 is true, or if 3 is true, or if 2 is true, or if 1 is true then..."

superbuggy:
otherwise good work!

starter/script.js
```diff
@@ -1 +1,74 @@
...
+      myFunction();
+      parseInt(x);
+      var ex = x.toString();
+  }
+else if (isNan(Unit)) {
+  alert("WRONG maggot! Try that again");
+  count ++;}
+else {
+  alert("sorry we've fucked up");
+}
```
superbuggy, line 74: I could not get your program to run. I know it is tempting, as someone with a CS background, to write a bunch of code without testing it. You've said that you are getting out of this habit so that is good. There were a lot of syntactical issues with your code, though. Just write a little bit and test it. It's not flashy, but it is _effective_.

starter/script.js
```diff
@@ -1 +1,74 @@
...
+    alert('Congrats! You honored your commitment to these US armed forces in support and defense of international law. Your unquestioned valor and courage \n
+    may have saved the Earth from sure destruction. You were placed in the Intel unit that uncovered the single alien vulnerability.');
+  }
+}
+
+else if (Unit == 'E3') {
+  alert(
+  ('Youre not new to this huh? I need you to prove your skills. press the button below to shoot the target. Youll need to land 5 shots to successfully enter this unit.'));
+  var skills = 0;
+  function shoot()
```
superbuggy, line 45: take this function out of the conditional; functions won't be defined in the global/top-level scope.

starter/script.js
```diff
@@ -1 +1,74 @@
...
+  {
+    while(shots<5)
+    {
+      shots += 1;
+      skills +=1;
+    }
+      document.getElementById("shots").innerHTML = shoot;
+    if (shots == 4)
```
superbuggy, line 53: Inventive implementation of the number requirement! 👍

PR #203, opened 2016-07-05
Van's Choose Your Own Adventure Game

VansPhan:
**Comfort:** 5
**Completeness:** 4

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,193 @@
-// put code here!
+var character = {
+	name: null,
+	inventory: [],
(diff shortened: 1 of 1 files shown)


nayana487:
Hey Van, 

Super awesome job with this! You went way above and beyond the scope of this assignment. My only thought is that you avoid using switch statements--it's cool that you took the time to learn about them--but it's preferable to use an object literal instead. Check out this article for more on that: https://toddmotto.com/deprecating-the-switch-statement-for-object-literals/

PR #205, opened 2016-07-05
Rasheeda A Choose Your own Adventure

RasheedaAlexander:
Completeness: 3
Comfort: 3

Comments: Oh, I struggled. My if, else statements produced the wrong results, I'm sure my variables were all over the place.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,0 +1,37 @@
+var choice = "";
+var type = prompt("You're finally old enough to choose your first Pokemon! Professor Oak gives you 3 pokeballs to choose from. What type of Pokemon do you choose? Water, Fire or Grass (W, F, G)?");
+if (choice == "W" || "Water") {
+  alert("Your first Pokemon is Squirtle!");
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -0,0 +1,37 @@
+var choice = "";
+var type = prompt("You're finally old enough to choose your first Pokemon! Professor Oak gives you 3 pokeballs to choose from. What type of Pokemon do you choose? Water, Fire or Grass (W, F, G)?");
+if (choice == "W" || "Water") {
+  alert("Your first Pokemon is Squirtle!");
+}
```
andrewsunglaekim, line 5: I already talked to you about this, but the string `"Water"`  is truthy, therefore no matter what the user inputs it will always return true and the first condition will always be met.

andrewsunglaekim:
looks fine except for the part i commented on. You fix that and alot of your functionality will start to work. Good work 👍

PR #206, opened 2016-07-05
Mike's adventure game

michaelmahony:
Comfort: 4,
Completeness: 4

The framework is in place and works well. A good next step would be to introduce the ability to add options within the scene as the user interacts with prompts. This could be done by creating a mapping between elements within the scene and elements that they unlock. Main can invoke a function to move inactive prompts into an active prompt list as the items that they are mapped to are interacted with.

Additional scenes can be added trivially, although as things are currently implemented, scenes must be uniquely named.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,77 @@
 // put code here!
+var situations = {
+  start : {
+    // The main text prompt of the scene
+    text: "It's eight o'clock on Thursday morning and you don't feel very good. You wake " +
+     "up blearily, wander around your room, opened a window, saw a bulldozer, found your " +
+     "slippers, and stomped off to the bathroom to wash.",
+
+    // A list of interactable objects to be shown to the user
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,115 @@
...
+
+      } else {
+        console.log("Input was a scene.");
+        user.set_new_scene(response);
+      }
+    }
+  }
+}
+
+main();
```
superbuggy, line 115: I like the encapsulation-based approach quite a bit! I also like how when you try an option that doesn't lead to another scene, it displays the outcome of that option-attempt in the alert, ala old text adventure games. I wasn't able to get out of the taunt/reason scene however. 

The assignment requirements specified a different event-tree than the one you implemented; not a huge deal but it's important to try to impose the constraints on yourself offered by the assignment. This is to model how a client or team might specify how to code a certain feature or functionality.

I think an encapsulated object style solution to this exercise can be somewhat tricky. Great work overall however!

PR #230, opened 2016-09-15
adventurehw

bkeating55:
complete -3
comfort - 3

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,35 @@
 // put code here!
+var yards = i;
+
+confirm("Are you ready for some football!?")
+var team = prompt("What's your favorite team?");
+  if (team == "Redskins") {
+    console.log("You are cool")
+  } else {
+    console.log("Better not be the Cowboys")
+  }
+  console.log("Welcome to Monday Night Football! Tonight the Redskins play the Cowboys");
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,35 @@
 // put code here!
+var yards = i;
```
nolds9, line 2: `i` is not defined here yet

starter/script.js
```diff
@@ -1 +1,35 @@
...
+    console.log("You lost the toss!")
+  }
+var result = prompt("Did you win the toss?");
+  if (result == "Yes") {
+    console.log("Congrats! You have choice on if you want the ball first or second half")
+  } else {
+    console.log("The other team has choice")
+  }
+ prompt("The Cowboys kick the ball to the Redskins")
+ prompt("How many yards did the Redskins get on kickoff")
```
nolds9, line 25: need to save user input to a variable

starter/script.js
```diff
@@ -1 +1,35 @@
...
+
+ for (var i = 0; 0 < 100; i++) {
+  if (yards = 100) {
```
nolds9, line 28: remember the difference between `=` and `===`

nolds9:
👍

PR #234, opened 2016-10-10
Sam Mans HW

sdmans:
comfort 2
completeness 2

Comment: I couldn't get the switch to work properly, it would console log all there statements regardless of the user input.


superbuggy:
👍

PR #238, opened 2016-10-10
Thinh's Choose Your Own Adventure

nguyendthinh:
Comfort: 5
Completeness: 4.8

I'm not really sure how to incorporate a loop in this kind of "gaming" concept.


starter/script.js
```diff
@@ -1 +1,78 @@
...
+        alert("Squidward begrudingly accepts your offer, at least you pay more than the Krusty Krabs");
+      } else {
+        alert("One of your responses was not valid")
+      }
+    break;
+
+    default:
+    alert("Please type a valid response");
+  }
+}
```
superbuggy, line 78: I thought it was interesting approach that you wrapped a bunch of conditionals inside of a switch, although stylistically this is not preferable. You could instead break the conditional statements out into separate functions and call those functions accordingly. Great work overall! 👍

superbuggy: Also using `toUpperCase()` to condition user input is awesome

PR #241, opened 2016-10-10
Juliet Berman Harry Potter Adventure

julietberman:
Completion: 3
Comfort: 3-4

My if/else statements work correctly individually, but I couldn't get them to run in order! I layered them inside each other, but it keeps looping back to the first question. Would love to go over how to kill functions after they have been run once. Comfort with storing the data to be reused and manipulating the DOM, not so much on how to use functions in JS.


andrewsunglaekim:
The return keyword will stop execution of a function. Additionally, make each part of the story it's own function. Then invoke those other functions at certain forks in your story. 👍

PR #242, opened 2016-10-10
Drew's Hw

drewswinburne:
Comfort: 4.5
Completeness: 4

I was unsure of how to incorporate a loop into this kind of game. Otherwise this was really fun to do


nayana487:
could be some kinda countdown on number of drinks consumed and activities available to do?

PR #243, opened 2016-10-10
Donovan's HW - CYOA

donovanallen:
Complete: 4
Comfort: 4
Comments: Had trouble with the switch and loop statements - hoping to get some more guidance on Monday.


andrewsunglaekim:
don't worry too much about switch statements, they're just alternate syntax for if/else conditionals. Definitely get up to speed on looping though! Let me know if you have questions about loops 👍

PR #245, opened 2016-10-10
Vanessa's_js-adventure

VanessaNya:
comfort: 3
completeness: 3.8

I'm not sure if I was on track here.  I tried using the loop and switch statement, not sure if I implemented them correctly.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,49 @@
 // put code here!
+confirm("I am ready to play!")
+var age = prompt("What's your age?");
+if (age is less than 13) {
+    console.log("you're underage, still want to play?");
+} else {
+    console.log("play it safe!");
+}
+console.log("You are stuck in the woods.'");
+console.log("Suddenly, you hear a growling noise'");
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,49 @@
...
+} else {
+  console.log("begin tracing your path for rescue");
+}
+console.log("you see a clue");
+var userAnswer = prompt("do you pick pick it up or keep tracing?");
+if (userAnswer = "yes") {
+  console.log("congrats! you're almost there");
+} else {
+  console.log("you have a long way to go");
+}
```
superbuggy, line 49: This a good start, but you'll need to nest coniditionals or break out the different choices into functions and then call those functions at the appropriate times to fulfill the criteria of the assignment. 

I think your switch will never log "lions roaring" because `growlingNoise` is always going to be "bear"

otherwise 👍 on camelCase variables and using `prompt()` and storing the result in a variable, indentation practices, and well-formed conditionals

PR #259, opened 2017-01-22
CHOOSE YOUR OWN ADVENTURE HW / SPACES / ISAAC DUARTE

Philocoder93:
Comfort: 3.5/5
Writing that many nested if/else statements made my head hurt, I can't help but think that there's an easier way to do that, I just wasn't able to come up with it yet.
Completeness: 5/5 
Unless I missed something I think this one is entirely complete.

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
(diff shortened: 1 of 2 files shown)


nayana487:
could use switch statements. 👍

PR #262, opened 2017-01-25
CYOA - "Pokemon Adventure" - Erica Norby

ericanorby:
Comfort level: 4
Completeness: 5

Started out copying and pasting different responses for each condition, then decided to condense it all into a function. It was challenging but fun. I might update with edits to the "while" loop so that the  evolution doesn't happen over and over again if the user keeps training.


phantomhaircuts:
👍

PR #264, opened 2017-01-26
Bedig Galladian | Create Your Own Adventure

bgalladian:
Comfort-3/5-I still need a lot of practice with Java.  Had to google a lot of different codes to realize what I should do.  I can read it fine, but creating is a whole new issue.

Completeness-4/5.  I feel like if I said 5/5 I would be lying, as I feel something is missing.  I just can't pinpoint what at the moment.

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -2,9 +2,10 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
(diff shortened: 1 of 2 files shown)


AndyWhitley:
👍

PR #265, opened 2017-01-26
Nat Carter | Choose Your Own Adventure

ghost:
Comfort: 5 / 5
Completeness: 5 / 5

I went way down the rabbit hole on this. I decided not to go with a pure if/else-if/else/switch approach, since I couldn't imagine how this would handle bad input. So my first real prototype comprised a function declaration for each event and one function call to the first event. Each function ran user input through a switch statement to call the appropriate next function, or to call itself if the user input was bad. This worked great, and it looked pretty sleek in Atom too, but all those functions repeated a lot of the same code as one another. What if I wanted to change something? I would have to change it in several different places! I wanted something DRYer. After tens of hours, I finally ended up with this.

Was it worth all that time? Maybe not? But it taught me a lot about hoisting, object literals, and dot versus bracket notation. It also drove home "commit early, commit often," since this is exactly the kind of case where having a history of my progress might be useful.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,8 +1,9 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-  <meta charset="UTF-8">
+  <meta charset="utf-8">
   <title>Document</title>
+  <link rel="stylesheet" type="text/css" href="style.css">
 </head>
 <body>
   <script src="script.js"></script>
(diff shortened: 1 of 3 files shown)


amaseda:
:+1: Yes to everything you said. Great job with this. Let me know if you have any lingering questions / comments on how you refactored / could keep refactoring this.

PR #267, opened 2017-01-26
CYOA HW - Vince Campanale

vincecampanale:
Completeness: 5
Comfort level: 5
Question: I think I read somewhere that `switch` statements are bad practice because they create their own scope or something...is that true and if so, what is the reasoning behind that opinion?

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>Document</title>
+  <title>Choose Your Own Adventure</title>
 </head>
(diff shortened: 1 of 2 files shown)


amaseda:
:+1:

Hm, I haven't heard of them being bad for that reason (although if you've got some resources on that I would def be interested in reading them).

The most common complaints I've seen have to do with readability. Having to write `break` multiple times sucks too.

Personally I don't find myself writing them that often. Also, [this is a nice alternative](https://javascriptweblog.wordpress.com/2010/03/08/caseagainstswitch/).

PR #271, opened 2017-01-26
Will Hawkins Epic Adventure

hawkins-will:
Completion: 5/5
Competence: 5/5

There may or may not be a Chipmunk involved


hawkins-will:
I want to point out that when encountering bandits, you can select the letter option OR input the monetary value. YOU CAN ENTER NUMBERS AT THAT POINT

nayana487:
this is incredible. 💯

PR #277, opened 2017-01-27
Nyketha's Static Shock Adventure

Nykke:
Level of Completeness: 5
Level of Understanding: 4

Comments: this was tough to think so logically, I couldn't figure out how to look all the variables. and I didn't do the parse.


starter/script.js
```diff
@@ -1 +1,115 @@
...
+var heroes = ['Batman', 'Static'];
+
+  for (i = 0; i < meta.length; i++);
+    if ('Ebon' == meta) {
+     console.log('Ebon loses to Static');
+   }
+      else if ('Talon' == meta) {
+        console.log('Talon loses to Batman');
+        }
+        else if ('Hotstreak == meta') {
```
superbuggy, line 108: careful with syntax here

`'Hotstreak == meta'` is a string, rather than an expression 
you want:
`Hotstreak == 'meta'` to use the comparison operator `==`. 
You have this correct elsewhere though 😃

superbuggy:
Good progress! 👍

PR #278, opened 2017-01-27
Collin O solution

cMoney667:
Level of Completeness: 5
Level of Understanding: 4.5

Had some issue with understanding and also having different errors that I had to work through were annoying but fell better now that I was able to trouble shoot them all.


phantomhaircuts:
👍

PR #285, opened 2017-01-27
Ben Shapiro HW - Choose Your Own Adventure

pretorben:
Comfort: 4
Completeness: 4

The program works for one play through, but I would have liked to be able to get it to reset so the user could play again if they wanted.  If I can figure out how to do that I'll resubmit.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,97 @@
-// put code here!
+var alive = 1;
+while(alive===1){
+  var response = prompt('Welcome to Choose Your Own Adventure: Harry Potter\n\nA: You are Harry Potter, walking down a dark corridor at Hogwarts with Hermione and Ron.  You are searching for the final Horcrux.  There is a door on your left and you hear a noise behind it.  Do you open the door? (y/n)');
+  switch(true){
+    case (response === 'y' || response === 'Y'):
+      var responseB = prompt('You open the door.  Crabbe and Goyle are waiting inside; Crabbe abducts Hermione and Goyle abducts Ron.  Do you chase after Hermione, Ron, or continue searching without them (h/r/n)?');
+        if(responseB === 'h' || responseB === 'H'){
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,97 @@
-// put code here!
+var alive = 1;
+while(alive===1){
+  var response = prompt('Welcome to Choose Your Own Adventure: Harry Potter\n\nA: You are Harry Potter, walking down a dark corridor at Hogwarts with Hermione and Ron.  You are searching for the final Horcrux.  There is a door on your left and you hear a noise behind it.  Do you open the door? (y/n)');
+  switch(true){
```
superbuggy, line 4: there are some interesting unconventional uses of switch for sure, but I just want to mention that with switch it is more typical to pass some variable in as the argument to switch. One of the reasons for this is that switch is particularly well-suited to considering three distinct cases where only one will be true.

reference: http://stackoverflow.com/questions/2765981/javascript-switchtrue

starter/script.js
```diff
@@ -1 +1,97 @@
...
+              alert('Your hug distracts you from seeing that Ron is under the Imperius Curse.  He unwittingly casts Avada Kedavra on both of you.  So much for stopping Voldemort!');
+              alive++;
+            }
+            else if(responseD === 'k' || responseD === 'K'){
+              alert("Remus Lupin is transformed into a werewolf and kills you as you're distracted by inexplicably kissing your best friend's girlfriend.  You jerk.");
+              alive++;
+            }
+            else{
+              alert('Invalid response.  Voldemort does not like invalid responses.  Avada Kedavra!');
+              alive++;
```
superbuggy: alive-- here?

starter/script.js
```diff
@@ -1 +1,97 @@
...
+          alive++;
+        }
+      break;
+
+    default:
+      alert('Invalid response.  Voldemort does not like invalid responses.  Avada Kedavra!');
+      alive++;
+      break;
+  }
+}
```
superbuggy, line 97: Nice work Ben! 👍

PR #286, opened 2017-01-27
HW-Adventure-Habtamu Abele

HabRonan:
comfort-4.5
completeness- 4.5

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,88 @@
 // put code here!
+//$("button").on("click", function(){
+//    console.log($(this).html());
+//  })
+
+
+
+/*document.getElementsByClassName("myBtn").addEventListener("click", displayThis);
+
+function displayThis() {
+    document.getElementById("display").innerHTML = "Great Choice,let's go.";
+
+    $(document).ready(function(){
+        $("#myBtn").click(function(){
+            $("#myBtn").hide();
+        });
+      });
+}
+*/
+
+
+
+var planet = $("button").on("click", function myFunction(){
+  document.getElementsByClassName('button').innerHTML
+  return ;
+});
+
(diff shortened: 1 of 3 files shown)


starter/script.js
```diff
@@ -1 +1,92 @@
...
+                          alert("your body can't survive this")
+                        break;
+                   case ("saturn"):
+                          alert("your body can't survive this")
+                          break;
+                   case ("neptune"):
+                        alert("your body can't survive this")
+                          break;
+
+                         }
```
amaseda, line 89: What happens if this switch statement is triggered but the user input does not match any of the planets you have specified?

starter/script.js
```diff
@@ -1 +1,92 @@
...
+        alert("you can't go back to the past unless you have a time machine!")
+      }
+
+}
+
+if (planet == "pandora"){
+        var type = prompt("great Choice, what kind would you like to be? a human, an avatar or a na'vi")
+        var check = ["human", "Avatar", "na'vi" ];
+        var text = " ";
+        var i;
```
amaseda, line 51: With for loops you do not have to define `i` prior to the loop (unless you have a specific reason for doing so). Instead, this is a common practice...


```js
for(var i=0; i<arr.length; i++){
  // code goes here
}
```
> Note that we are initializing `i` inside `for`

amaseda:
:+1:

starter/script.js
HabRonan: ya, I had a default statement but that one gets executed all the time and then I took it out.
I guess I have to look into that.

starter/script.js
HabRonan: Alright, Thanks.

PR #288, opened 2017-01-27
Choose Your Own Adventure HW (Justin Shin)

Justin-Shin:
Completeness: 4.5
Comfort: 3

I say completeness at 4.5 because I ask numerous questions, but some felt like a cheat to get questions out there.  With that in mind, I could have made more prompts, but I didn't want to make it too big because I was trying to make a sort of framework for the game that was a bit more scalable.  I don't think I succeeded particularly well, but it beats the heck out of using nested if statements.

I also am terrible at commenting in code and in the future will try to make sure I add documentation.

I put a comfort level of 3 because while I eventually succeeded, I coded this on the fly and approached it in a different way.  As mentioned before I wanted it to be scalable for bigger adventures so to speak, and I could easily add extra prompts in the object.  Basically I created a crappy JS database by forming an object populated by arrays.  I let the first element in each array dictate the way the program would express each array- that is it would expect some number of potential responses based on what the identifier was.  Then, in the array I can pull the element that corresponds with the response and set the new prompt identifier, and run through my do while loop.

Anyway, I can just add arrays to the object for more prompts if needed, although it would require some tweaking.

One thing I couldn't figure out but hacked together was an ending parameter- that is how to tell when to go to an ending message.  I made the last element in the arrays act as a identifier, and the do while loop checked to see if it get flipped.  The way to flip that element was in a function I made that is run every time a prompt is complete.  It checks to see if the response given and the prompt id matches, and if it does toggles it.  This was a very clunky way, and I would have to hardcode each combination in the future, which I don't like.  Hopefully there's a better solution for this somewhere.

Also, if the user says to stop showing prompts in the middle of the adventure, my script will keep running and attempting to prompt the user with the same question if they say cancel.  It can cause the browser to freeze.  I would try to figure out a good way to quit here, but I'm sleepy and have to go over my other homeworks to make sure they're done!

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,136 @@
-// put code here!
+console.log("Hello World");
+alert("Hello, you are a Hobbit named Bilbo that lives in the Shire!");
+alert("A lifetime of adventure could await you...if you would like.");
+alert("Possible responses are provided. just enter the first letter of the response you want to choose!");
+var userName = window.prompt("Bilbo, what is your name?");
+var friend
+while(userName != "Bilbo") {
+  userName = window.prompt("Err, are you sure your name isn't Bilbo?")
+  console.log(userName);
+}
+var currentPrompt = ageQuery();
+var questionType = 1;
+var promptMap = {
+  "continue01": [1,"Brilliant! Chronicle your life in your new book, The Hobbit, to there and back again!  Remember Smaug? Oh those were the days...","prompt03",0,null,0,null,0,null,0],
+  "continue02": [1,"I don't know half of you half as well as I should like; and I like less than half of you half as well as you deserve.' - You feel awkward! Time to vanish away. You slip on the ring and run off away from the party leaving everything to Frodo.","prompt04"],
+  "continue03": [1,"Of course we will invite "+friend+"Here they come with Gandalf!","prompt05"],
+  "continue04": [1,"How ever so adventurous of you.  I'd tell you the rest, but that's a story for another time.","prompt06"],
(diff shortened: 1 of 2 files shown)


starter/script.js
```diff
@@ -1 +1,136 @@
...
+    currentPromptHolder = "prompt09";
+  } else {
+    window.alert("Oh come now, let\'s try this again...");
+    currentPromptHolder = ageQuery();
+  }
+  console.log(currentPromptHolder)
+  console.log(ageHolder)
+  return currentPromptHolder;
+}
+function checkAnswerType(prompt,answer){
```
superbuggy, line 116: I like the thought behind this approach with your choice-validator function

starter/script.js
```diff
@@ -1 +1,136 @@
...
+  } else if (prompt=="prompt05" && answer == "D"){
+    return 1
+  } else if (prompt=="prompt08" && answer == "N"){
+    return 1
+  } else if (prompt=="prompt09" && answer == "B"){
+    return 1
+  } else {
+    return 0
+  }
+}
```
superbuggy, line 136: I like the thought behind your program architecture. I think it could really be simplified and made more elegant if you separate/decouple the map of the flow of events from the questions (what you have in `promptMap`). It's good to keep your data-structures well-organized and as simple as possible. Using `true` and `false` is preferable to using 1 and 0 as well, since they are more semantic. 

The robust effort and thought behind this assignment is very admirable and as is your willingness to tack complexity.

PR #290, opened 2017-01-27
nobody expects the spanish inquisition

xgironx:
complete:  4, would like to have fixed some pieces, but i think it meets the reqs
comfort:  moreso than the DOM, for sure.


starter/script.js
```diff
@@ -1 +1,124 @@
...
+      anotherTripVarText = promptLetsStart + '\n' + prompt11SpyNameVar;
+      // alert(anotherTripVarText);
+      break;
+    case 'N':
+      text = 'Stay at home, drink alone!';
+      break;
+  // for each(var x in adventure) {
+  //   console.log(adventure.prompt20UglyAmericanStatus[x]);
+  // }
+}
```
superbuggy: Your architectural approach is ambitious, but it might be simpler to debug starting out with a simpler program architecture. It hit an error at line 87 (toUpperCase is a method, ex `someString.toUpperCase()`). Great thought and effort overall, though. Start simple!

PR #306, opened 2017-03-29
Mario

zonekid:
Gave me headaches!!! But once I got it I understood quite clearly

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,64 @@
-// put code here!
+var user = prompt ("Would you like to play mini Mario. Type Yes or No");
+
+while (user == "Yes") {
+  var play = prompt ("Ok great! Toad has a message for you. Hi Mario! You have three options at your disposal to defeat Bowser and save the Princess. Type 1 for Hammer, Type 2 for Mushroom, or Type 3 for Yoshi's Egg" )
+
+  switch(parseInt(play)) {
+    case 1:
+      weapon = "Hammer"
+      break;
+
+    case 2:
+      weapon = "Mushroom"
+      break;
+
+      case 3:
+      weapon = "Egg"
+      break;
+
+      default:
+      text = "";
+
+  }
+  var guard= prompt (" You have encountered the guard of Bowser's castle. Fight or Make Peace?")
+    if ("fight" == guard) {
+        var entry= prompt ("You've succesfully defeated the guard but Bowser is missing from the castle. Would you like to find him (type find) or just wait for him to get back? (type wait)")
+    } else {
+        var peace= alert ("Sorry but Super Heros have to fight! Game Over!")
+    }
+            if ("find" == entry) {
+                var star= prompt("You are now in the Star Room. Theres a huge talking star near the apple tree. He says that Bowser awaits at Rainbow Road. Go to Rainbow Road (type rainbow) or ignore the star(type ignore) and proceed")
+            } else {
+                var wait= alert("It seems Bowser used your time of waiting to slay off the Princess.  Game Over!")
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,64 @@
...
+
+      default:
+      text = "";
+
+  }
+  var guard= prompt (" You have encountered the guard of Bowser's castle. Fight or Make Peace?")
+    if ("fight" == guard) {
+        var entry= prompt ("You've succesfully defeated the guard but Bowser is missing from the castle. Would you like to find him (type find) or just wait for him to get back? (type wait)")
+    } else {
+        var peace= alert ("Sorry but Super Heros have to fight! Game Over!")
```
juancgarcia, line 27: `alert()` doesn't supply any return value so trying to assign it to a variable doesn't really make sense. Also the `peace` variable isn't used anywhere else in the code so there's no need to declare this variable.

starter/script.js
```diff
@@ -1 +1,64 @@
...
+    }
+            if ("find" == entry) {
+                var star= prompt("You are now in the Star Room. Theres a huge talking star near the apple tree. He says that Bowser awaits at Rainbow Road. Go to Rainbow Road (type rainbow) or ignore the star(type ignore) and proceed")
+            } else {
+                var wait= alert("It seems Bowser used your time of waiting to slay off the Princess.  Game Over!")
```
juancgarcia, line 32: `alert()` has no return and variable `wait` isn't used

starter/script.js
```diff
@@ -1 +1,64 @@
...
+            }
+
+                   if ("rainbow"== star){
+                      var brave= prompt("You have made it to Rainbow Road.  Theres a magic carpet ride going down a dark space.  Be brave and take the ride (type ride) or play it smart? (type smart)")
+                } else {
+                    var ignore= alert("The star is very wise.  You did not seek his wisdom.  You are now lost and Bowser has done away with the Princess! Game Over!")
```
juancgarcia, line 38: `alert()` has no return and variable `ignore` isn't used

starter/script.js
```diff
@@ -1 +1,64 @@
...
+                }
+
+                        if ("ride"== brave) {
+                            var Bowser= prompt("You have now found Bowser deep in his monster cave.  It's time for action!")
+                        }      else {
+                                  var smart= alert("Chicken! You'll never be able to find Bowser! Game Over")
```
juancgarcia, line 44: `alert()` has no return and variable `smart` isn't used

starter/script.js
```diff
@@ -1 +1,64 @@
...
+                              }
+
+                            switch(parseInt(Bowser)) {
+                              case 1:
+                                alert = "The hammer was sturdy enough to knock off the head of Bowser! Congrats! You've slayed the beast!"
```
juancgarcia, line 49: Remember, `alert` is defined as a function in the global scope. If you assign a value to `alert`, the function will be destroyed and can't be called anymore! Be careful of your variable name choices to avoid this in the future.

juancgarcia (commented):
There's quite a few problems here with your code in addition to the line items I noted.

In particular I think you're struggling greatly with understanding how code gets grouped for control flow.

Starting on line 29 `if ("find" == entry) {` you begin indenting each `if\else` one level further than the previous. It's not only confusing to read, but suggests you think that the indentation causes the indented lines to somehow be included in the previous `else` block. This is not the case.

Please review this link on your own as further reading on `if/else` statements.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Control_flow_and_error_handling

PR #307, opened 2017-03-30
Adventure HW - Erin Dawson

edawson1980:
Comfortability: 2
Completeness: 2

I've been working on this for two days, and everything just kind of fell apart on me.  I tried to redo different versions in different script files in the hopes that when I figured it out, I would just delete the unnecessary scripts.  
That didn't really work out.
I've clearly hit a wall here, and I don't understand what's not clicking.  It doesn't seem to matter how many times it's explained to me, or how much I google.
I understand if this homework counts against me.  I did honestly try, I'm not sure how to proceed.

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -2,9 +2,10 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>Document</title>
+  <title>choose your own adventure</title>
 </head>
 <body>
-  <script src="script.js"></script>
+  this is a game i wrote about the mysteries of the universe.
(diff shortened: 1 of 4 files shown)


jsm13:
👍 This totally counts. I noticed in a couple of the scripts, the returned value of `confirm` or `prompt` is not being stored. Whenever we make a decision based on some user interaction, we want to store the result in a variable so `var ready = confirm("are you ready?")`. We can then write logic like `if (ready) <something> else <something else>`. Without the variable assignment, the result of the `confirm`/`prompt` is forgotten

PR #310, opened 2017-03-30
Choose your own adventure Jeong Hoon Koo

jkoo87:


starter/script.js
```diff
@@ -1 +1,42 @@
-// put code here!
+var quit = false;
+var fullName = prompt("What is your name?");
+var result = prompt("Hello " + fullName +", are you ready to conquer the World?\n- Yes\n- No");
+if ("No" == result) {
+	alert("Have a good life");
+	quit = true;
+	window.location = "http://google.com";
+} else if ("Yes" == result) {
```
juancgarcia, line 8: Since this question only takes 2 answers, your `if` statement only needs to check for one. the `else if` can just be `else`.

The reason `else if` might create a problem is if your user types something other than "Yes" or "No" like "Maybe". That won't be covered by either `if`, none of the blocks get called and `yourCountry` is never assigned a value. For the rest of the game `yourCountry` is undefined.

Something like this could work better.
```js
if ("No" == result)
    alert("goodbye");
} else {
    var yourCountry = prompt("Choose a country");
}
```

juancgarcia (commented):
Good job.

PR #311, opened 2017-03-30
Abe's attempt

aeaghannam:
I was unable to make my operations react to any input, I will need to review the Js lessons in detail this week....still trying to understand the basic syntax by going through code academy and code camp demos online and reading.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,9 @@
-// put code here!
+confrim ("Hello are you ready to begin this adventure?!");
+
+var age = prompt("What's your age");
+
+if(age > 13) {
+  console.log("Welcome!"");
+}else {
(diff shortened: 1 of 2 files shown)


starter/script.js
```diff
@@ -1 +1,9 @@
-// put code here!
+confrim ("Hello are you ready to begin this adventure?!");
+
+var age = prompt("What's your age");
+
+if(age > 13) {
+  console.log("Welcome!"");
```
jsm13, line 6: The second `"` after welcome is causing a syntax error. That's going to cause the whole script to break

jsm13:
You need to be careful of syntax errors. That can derail everything. Be sure to always have the dev console open to check for errors

PR #312, opened 2017-03-30
HW (Meag Doherty)

meagdoh:
Comfort score, 1 out of 5
Completeness score, 1 out of 5


jsm13:
Make sure you are writing your code in a script file which is being loaded. First line of JS we write should always be a `console.log('here')` so you can check in the browser that the script is running

PR #313, opened 2017-03-30
WPurdum-Choose Your Own Adventure

wjpurdum:
Comfort level - 3
This was harder for me. It took me a while to understand how prompt works - originally I overcomplicated it and made a different var for each choice. 
- I couldn't figure out how to make my 'while' loop go back into the logic once a valid input was entered. 

- I couldn't get my switch statement to work. I need to use this more in my code to get how it works.

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -11,12 +11,28 @@ var year = prompt("Hello, " + username +  " Okay, let's select our destination. 
     } else var ninetyfour = prompt("Sorry, I didn't catch that. Where are we going? 1950/1965/1989/1994");
 
 
+//Options for first question for 1950//
 if (fifty === "T") {
 	var trumanOne = prompt("Excellent. President Truman would like to Prep for the meeting. Would you like to discuss domestic or international affairs? (D/I");
 } else if (fifty === "M" ) {
 	var mccarthurOne = prompt("Excellent. the General looks forward to the discussion. Would you like the meeting to take place on sea or on land? (S/L")
 } else {
 	prompt("Sorry, I didn't catch that. Press 'T' if you'd like to have lunch with President Truman. Press  'M' if you'd like to have lunch with Douglas MacArthur.")
+}
+
+
+
+//Options for Truman Question 1
+if (trumanOne === "D") {
+	var trumanD = prompt("Good choice - there is much to be discussed. Would you like to talk about the McCarran Internal Security Act, or the Kefauver Committee? (M/K");
+} else if (trumanOne === "I") {
+	var trumanI = prompt("Good choice - there is much to be discussed. Would you like to discuss the ratification of the North Atlantic Treaty Organization, or the Korean War? (N/K")
+}
+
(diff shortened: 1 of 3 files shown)


starter/script.js
```diff
@@ -1 +1,131 @@
-// put code here!
+var username = prompt("Hello there! You've been given a chance to go back in time to meet with some of modern history's most important political and popular figures. Before you meet them, you'll need to answer a few questions, and do some preperation for your meeting. Are you all set? What is your name");
+
+var year = prompt("Hello, " + username +  " Okay, let's select our destination. We have a few options. Enter the decade you'd like to visit: (1950/2030)")
+    if (year === "1950") {
+      var year = prompt("Heading to the 50s! Interesting time. Would you like to have lunch with President Harry Truman or Frank Sinatra? (T/S)")
+    } else if (year === "2030") {
+      prompt("Just kidding! This hasn't happened yet! Why not try 1950?")
+  } while (year != "1950" || "2030") {
+  	prompt("Please select a valid choice!")
```
juancgarcia, line 9: Good idea, forcing them to make a valid choice with the `while` loop, but you're never capturing their input! You need to explicitly capture the return value from prompt via assignment or it will be lost.

```js
year = prompt("Please select a valid choice!")
```

starter/script.js
```diff
@@ -1 +1,131 @@
...
+  }
+
+  
+//Options for first question for 1950//
+if (year  === "T") {
+	var year = prompt("Excellent. President Truman would like to Prep for the meeting. Would you like to discuss domestic or international affairs? (D/I)");
```
juancgarcia, line 15: Redeclaring the variable `year` several times with `var` will work, but it's considered cleaner to only use var once for each variable. This would be a good opportunity to declare all the variables at the top `var year;` and then each subsequent assignment can drop the `var` keyword.

starter/script.js
```diff
@@ -1 +1,131 @@
...
+	var year = prompt("Let the good times roll. Would you rather discuss the high points in his career, or in his personal life? (C/P)")
+} else if (year === "B") {
+	var year = prompt("How macabre! Would you rather discuss the low points in his career or his personal life? (C/P)")
+}
+
+if (year === "C") {
+	var year = prompt("To give you some context for your meeting with Frank, let's go see him perform at 1951 at the Las Vegas Desert Inn. Sinatra was one of Vegas's biggest stars in the 1950s.")
+} else if (year === "P") {
+	var year = prompt("To give you some context for your meeting with Frank, let's go talk with actor Gregory Peck, who partied hard with Frank in Vegas during these years. Enjoy!")
+}
```
juancgarcia, line 49: Be careful with indentation. At a glance, it looks like the preceding code is on the top level, but it's actually nested inside of the `else` block started on line 16. All of the lines inside of that pair of `{}` should be indented to visually indicate to anyone looking at your code that they are inside of that `else` block.

juancgarcia (commented):
Good work.

PR #314, opened 2017-03-30
Neverending Adventure

lgmusic24:
Based off of the Neverending Story, would you rescue the Empress or find the hidden treasure?

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,94 @@
 // put code here!
+var name=prompt('What is your name?')
+ // ask which quest to discover
+ var quest=prompt('What quest would you like to go on? Save the Empress or Find the treasure? (e/t) ')
+
+ if(prompt=="e"){
+  //  Directions to the mission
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,94 @@
 // put code here!
+var name=prompt('What is your name?')
+ // ask which quest to discover
+ var quest=prompt('What quest would you like to go on? Save the Empress or Find the treasure? (e/t) ')
+
+ if(prompt=="e"){
```
jsm13, line 6: want to check the value `quest` here so `if (quest === "e")`. We are using the prompt to ask the question, and then the response is being assigned to the variable `quest`

jsm13:
👍 Good work! One thing to keep in mind is that we want to build in very small chunks. Think of the smallest thing that constitutes forward progress (eg ask the user a question and print the answer) and get that working before adding a tiny bit more functionality (making a programatic decision based on the answer).

PR #315, opened 2017-03-30
Not Completed

pchristianbrown:
I spent WAY too much time trying to create a story that I did not complete the work in a timely manner. I will continue to work on it in my own time.


jsm13:
If you get a chance to implement some of the logic on this and want some feedback just let me know

PR #316, opened 2017-03-30
Final CYOA

cancanomar:
Completeness 5/5
Comfort level 4/5

I think there were some issues but it should be okay.

-Omar


AndyWhitley:
👍

PR #332, opened 2017-07-09
choose your own adventure - colleen

colleeno:
complete - 5
comfort - 4

*feedback welcome, couldn't figure out how to stop an extra 'undefined' pop up after clicking on last alert, ended up wrapping all in an immediately called function to be able to return out to stop it

### DIFF
--- a/starter/index.html
+++ b/starter/index.html
@@ -2,9 +2,10 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>Document</title>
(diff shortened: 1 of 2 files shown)


jsm13:
👍 Awesome! When I run the code, get rid of the IIFE and return statements, I get the expected behavior but there may be a condition I wasn't checking. Surprised that it was coming up `undefined` as you don't reassign anything to `incorrectOption`.

PR #333, opened 2017-07-09
Jacob Greenwald (adventure homework)

jacobg1:
comfort 5
completion 5


starter/script.js
```diff
@@ -1 +1,87 @@
-// put code here!
+function game () {
+  $('.startGame').remove()
+  var name = prompt('Enter Name')  // prompt user to enter name
+  $('h2').text(`Greetings ${name} welcome to the game`) // return greeting with name
+  choice1 = prompt(`press 1 for dragons, 2 for unicorns or 3 for really much bigger dragons`) // give user first choice of three different options
+
+  switch (choice1) {
+    case choice1 = '1':
```
juancgarcia, line 8: In a `case` statement, we only need to provide the value we're testing against. The variable we're checking `choice1` in this case is already defined in the `switch()` declaration so we don't need to repeat it. That means `'1'` instead of `choice1 = '1'`

starter/script.js
```diff
@@ -1 +1,87 @@
...
+      location.reload()
+    } else if (choice2a === 'b') {
+      alert('YOU DIED')
+      location.reload()
+    }
+  }
+  if (choice1 === '3') {
+    $('p').text('these dragons are way too big.')
+    choice3a = prompt('you come to a 4 way crossing, which way do you go? Enter F/B/R/L').toLowerCase()
+    switch (choice3a) {
```
juancgarcia, line 56: Again no need to re-use the variable name inside your `case` statements.

juancgarcia:
👍 Nice job!

PR #334, opened 2017-07-09
Alex W Game of Thrones

awandres:
complete - 4
comfort - 4

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,122 @@
 // put code here!
+var endgame = false
+
+while (endgame === false) {
+var name = prompt('What is your name?')
+
+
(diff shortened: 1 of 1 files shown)


starter/script.js
```diff
@@ -1 +1,122 @@
 // put code here!
+var endgame = false
+
+while (endgame === false) {
```
juancgarcia, line 4: This isn't wrong, but when we know that our variable will only be `true` or `false` , instead of `variable_name === false`, we can more simply test for `false` with `!variable_name`

juancgarcia (commented):
👍 Nice Job!

PR #336, opened 2017-07-10
Charles Cox Choose Own Adventure

cwcox:
completeness- 5
comfort- 5

### DIFF
--- a/starter/script.js
+++ b/starter/script.js
@@ -1,1 +1,301 @@
-// put code here!
+function welcome () {
+  var know = prompt(
+    `Hello there Boddy.
+
+    Just a few minutes ago you were enjoying a nice evening
+    with Professor Plum, Ms. Scarlet, Mrs. Peacock,
+    Colonel Mustard, Mrs. White, and Mr. Green.
+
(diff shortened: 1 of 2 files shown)


starter/script.js
```diff
@@ -1 +1,301 @@
...
+    }
+}
+
+function conservatoryWoman () {
+  var accused = prompt(`We're almost there! What weapon did the killer use?
+
+  Mrs. Peacock with the [Knife]?
+  Mrs. White with the [Rope]`)
+
+  if (accused === 'Knife' || 'knife') {
```
juancgarcia, line 125: It looks like you're trying to compare the variable `accused` to the strings `'Knife'` and `'knife'`, but this statement doesn't work quite the way you're intending.

Because of operator precedence, it will be evaluated first as `accused === 'Knife'` and then if that fails, the string `'knife'` will be evaluated as a boolean, which will be `true` like all non-empty strings.

In order to test a variable against two strings you'll need to use the `===` operator for each variable/string combination:

```js
accused === 'Knife' || accused === 'knife'
```

juancgarcia (commented):
👍 Good Job.

